PR #28, opened 2016-12-29
Property smart completion

fharms:
- Adding support for smart completion inside property files.
- Refactor CamelJavaReferenceContributor to use CamelContributor for re-use


src/org/apache/camel/idea/CamelJavaReferenceContributor.java
```diff
@@ -16,137 +16,22 @@
...
-            if (answer != null) {
-                return answer.toArray();
-            } else {
-                return Collections.emptyList().toArray();
-            }
-        }
+public class CamelJavaReferenceContributor extends CamelContributor {
+
+    public CamelJavaReferenceContributor() {
+        extend(CompletionType.BASIC,
```
davsclaus: I wonder if we need to filter a bit more like the old code where it was checking for PsiLiteral. However we still check for PisLiteralExpression later to know its a string value.

fharms: Actually I tried the first approach with PsiLiteral, but that didn't work for some reason. But I limit it by checking if psiElement().inside(PsiClass.class)

PR #38, opened 2016-12-30
Smart completion property placeholders

fharms:
Add support for smart completion of property placeholder inside java files and property files


src/org/apache/camel/idea/CamelContributor.java
```diff
@@ -117,4 +117,55 @@ public void addCompletions(@NotNull CompletionParameters parameters,
...
+                File file = new File(virtualFile.getPath());
+                Properties properties = new Properties();
+                try {
+                    properties.load(Files.newInputStream(file.toPath()));
+                } catch (IOException e) {
+                }//TODO : log a warning, but for now we ignore it and continue.
+
+                properties.forEach((key, value) -> {
+                            String keyStr = (String) key;
+                            boolean noneMatch = IGNORE_PROPERTIES.stream().noneMatch(s -> keyStr.startsWith(s));
```
davsclaus: Should this be noneMatch ? I wonder if you should not filter the keys, eg if a properties file have both ignored and non ignored keys then we want still to include the non ignored key/values.

src/org/apache/camel/idea/CamelContributor.java
```diff
@@ -117,4 +117,55 @@ public void addCompletions(@NotNull CompletionParameters parameters,
...
+                            if (noneMatch) {
+                                LookupElementBuilder builder = LookupElementBuilder.create(keyStr+"}}")
```
davsclaus: I wonder if we should put the name of the properties file somehow in that list if possible (eg in the left aligned, where the regular one have the type name today).

src/org/apache/camel/idea/CamelContributor.java
fharms: yes, which mean none of the keys that is matching the filter list. 

The filter is only on key level, so any non ignore keys in a property file will still be included.

src/org/apache/camel/idea/CamelContributor.java
fharms: I thought about it myself, but the list just got to verbose and big. But I can try to make a version with the filename?

src/org/apache/camel/idea/CamelContributor.java
davsclaus: Ah doh yeah that is right

src/org/apache/camel/idea/CamelContributor.java
davsclaus: Yeah we can try later, it can just be the name [foo] or [foo.properties] but maybe the latter is too big as you say. Lets first get this baby in.

PR #52, opened 2017-01-01
Smart completion yaml

fharms:
- Adding support for YAML properties


src/main/java/org/apache/camel/idea/CamelContributor.java
```diff
@@ -21,110 +21,50 @@
...
-                }
-
-                // are there any results then add them
-                if (answer != null && !answer.isEmpty()) {
-                    resultSet.withPrefixMatcher(val).addAllElements(answer);
-                }
-            }
+            camelCompletionExtensions.stream()
+                    .filter(p -> p.isValid(parameters,context,tuple))
+                    .forEach(p -> p.addCompletions(parameters,context,resultSet,tuple));
```
davsclaus: I think we should not call multiple, eg it should be first found that is invoked only.

src/main/java/org/apache/camel/idea/completion/extension/CamelPropertyCompletion.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.apache.camel.idea.completion.extension;
+
+import com.intellij.codeInsight.completion.CompletionResultSet;
+import com.intellij.openapi.vfs.VirtualFile;
+
+/**
+ *
+ */
+public interface CamelPropertyCompletion {
```
davsclaus: Add javadoc to this interface

src/main/java/org/apache/camel/idea/completion/extension/YamlPropertyPlaceholdersSmartCompletion.java
```diff
@@ -0,0 +1,127 @@
...
+            result = (Map<String, Object>) yaml.load(ios);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        return result;
+    }
+
+    @Override
+    public boolean isValidExtension(String filename) {
+        return filename.endsWith(".yaml");
```
davsclaus: It can also be .yml for yaml files

src/main/java/org/apache/camel/idea/CamelContributor.java
fharms: Sure we can change it. But it limits to only one extension can decorate the list and you might have multiple sources you need to read from to decorate the list in some cases

src/main/java/org/apache/camel/idea/completion/extension/CamelPropertyCompletion.java
fharms: Will do

src/main/java/org/apache/camel/idea/completion/extension/YamlPropertyPlaceholdersSmartCompletion.java
fharms: damn it.... What happen to plain properties :-)

fharms:
Updated the Pull-request

src/main/java/org/apache/camel/idea/CamelContributor.java
davsclaus: Ah sorry users can have for example both properties and yaml files which we should support both

davsclaus:
For showing the file name in the list, that is nice. I have seen other plugins show this with hard brackets such as `[foo.properties]` instead of the double colon you use.

Also maybe try to set it as the _java type_ then its right aligned and maybe looks nicer?

src/main/java/org/apache/camel/idea/completion/extension/YamlPropertyPlaceholdersSmartCompletion.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public boolean isValidExtension(String filename) {
+        return filename.endsWith(".yaml") || filename.endsWith(".yml");
+    }
+
+    @Override
+    public void buildResultSet(CompletionResultSet resultSet, VirtualFile virtualFile) {
+        getProperties(virtualFile).forEach((key, value) -> {
+            final String keyStr = key;
+            boolean noneMatch = IGNORE_PROPERTIES.stream().noneMatch(s -> keyStr.startsWith(s));
```
davsclaus: This can be optimized as you build a stream for every single key/value pairs for all the properties there is.

Instead the old code I did with the array[] would be much faster.

fharms:
Good idea. Will try that.

Den 1. jan. 2017 6.29 PM skrev "Claus Ibsen" <notifications@github.com>:

src/main/java/org/apache/camel/idea/completion/extension/YamlPropertyPlaceholdersSmartCompletion.java
```diff
@@ -0,0 +1,130 @@
...
+                        buildResultSetForList(resultSet,virtualFile,keyStr, (List<?>) e);
+                    } else if (e instanceof LinkedHashMap) {
+                        buildResultSetForLinkedHashMap(resultSet,virtualFile,keyStr,propertyList);
+                    }
+                });
+    }
+
+    private void buildResultSet(CompletionResultSet resultSet, VirtualFile virtualFile, String keyStr,String value) {
+        LookupElementBuilder builder = LookupElementBuilder.create(keyStr + "}}")
+                .appendTailText(value + " :: " + virtualFile.getPresentableName(), true)
```
davsclaus: Use `[filename]` syntax for the filename. And maybe try to st it as the java type instead so its right aligned.

src/main/java/org/apache/camel/idea/completion/extension/YamlPropertyPlaceholdersSmartCompletion.java
```diff
@@ -0,0 +1,130 @@
...
+        Map<String, Object> result = new HashMap<>();
+        File file = new File(virtualFile.getPath());
+        Yaml yaml = new Yaml();
+
+        try {
+            InputStream ios = new FileInputStream(file);
+            // Parse the YAML file and return the output as a series of Maps and Lists
+            result = (Map<String, Object>) yaml.load(ios);
+        } catch (Exception e) {
+            e.printStackTrace();
```
davsclaus: We should probably just ignore this error. Or find out how to use IDEA logger and then log to idea log instead.

src/main/java/org/apache/camel/idea/completion/extension/YamlPropertyPlaceholdersSmartCompletion.java
fharms: I have refactored the code

PR #65, opened 2017-01-02
Intellij plugin test

fharms:
This one was a tough one to crack but we now have IntelliJ plugin testing working both from IDEA and Maven.

Additional test cases will be added in another pull-request


fharms:
Btw you will need to run the ./install-intellij-libs.sh again to add the properties.jar to the local maven

Issue #35, opened 2016-12-29
Add Camel icon in gutter


davsclaus:
Wonder if we should add a Camel icon on the lines which has detected a Camel endpoint? 
For spring beans you also have an icon in the gutter when its a Spring Bean / @Bean or something.

PR #95, opened 2017-01-05
#35 Added Camel icon in the gutter

adrianbumbas:


src/main/java/org/apache/camel/idea/completion/extension/CamelRouteLineMarkerProvider.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collection;
+
+/**
+ * Provider that adds the Camel icon in the gutter when it detects a Camel route.
+ */
+public class CamelRouteLineMarkerProvider extends RelatedItemLineMarkerProvider {
+    @Override
+    protected void collectNavigationMarkers(@NotNull PsiElement element,
+                                            Collection<? super RelatedItemLineMarkerInfo> result) {
+        if (element instanceof PsiLiteralExpression) {
```
fharms: nice work. remember to check if camel is present 
if (ServiceManager.getService(element.getProject(), CamelService.class).isCamelPresent()) {

src/main/java/org/apache/camel/idea/completion/extension/CamelRouteLineMarkerProvider.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Provider that adds the Camel icon in the gutter when it detects a Camel route.
+ */
+public class CamelRouteLineMarkerProvider extends RelatedItemLineMarkerProvider {
+    @Override
+    protected void collectNavigationMarkers(@NotNull PsiElement element,
+                                            Collection<? super RelatedItemLineMarkerInfo> result) {
+        if (ServiceManager.getService(element.getProject(), CamelService.class).isCamelPresent()) {
+            if (element instanceof PsiLiteralExpression) {
```
davsclaus: We have IdeaUtils where there is a method to get the text value of an element. This supports all the other languages like scala etc. So favor using that

src/main/java/org/apache/camel/idea/completion/extension/CamelRouteLineMarkerProvider.java
```diff
@@ -0,0 +1,57 @@
...
+                PsiLiteralExpression literalExpression = (PsiLiteralExpression) element;
+                String value = literalExpression.getValue() instanceof String ? (String) literalExpression.getValue() : null;
+                String componentName = StringUtils.asComponentName(value);
+                Project project = element.getProject();
+                if (value != null && !value.endsWith("{{") && componentName != null
+                        && ServiceManager.getService(project, CamelCatalogService.class).get().findComponentNames().contains(componentName)) {
+                    NavigationGutterIconBuilder<PsiElement> builder =
+                            NavigationGutterIconBuilder.create(IconLoader.getIcon("/icons/camel.png")).
+                                    setTargets(element).
+                                    setTooltipText("Camel route");
```
davsclaus: Dont you set an icon for every Camel endpoint? So its more Camel endpoint than Camel route?

If you want to detect a Camel route, then it would be on the class level, if the Class extends org.apache.camel.builder.RouteBuilder. Or we can go a little deeper, and find where a new route starts using the from method.

adrianbumbas: Yes, it is one icon per endpoint. The user can click on the gutter icon to do some action, depending on that action you can decide if you want the icon at the route level or endpoint level.

src/main/java/org/apache/camel/idea/completion/extension/CamelRouteLineMarkerProvider.java
adrianbumbas: I will check this class.

fharms:
@adrianbumbas FYI I mention you on my latest release tweet with name. Could't find your twitter name https://twitter.com/fnharms/status/817371987748200449

adrianbumbas:
Thank you for mentioning my name. Unfortunately I don't have a Twitter account.

PR #107, opened 2017-01-06
Camel is present polished

fharms:
- Be more smart about when we instantiate the catalog
- Add initial test case for testing documentation
- Optimize the memory consumption in CamelService
- Optimize performance in CamelAddEndpointIntention.java


fharms:
It seem that it does not clean up after tearDown and keep the instance of the CatalogService

fharms:
c'mon....

davsclaus:
@fharms I can see you have a bit of fun.

Great work finding out about how to make the test robust and the dispose thingy.

Issue #109, opened 2017-01-06
The gutter icon also appears for non-Camel code


adrianbumbas:
The gutter icon shows up for every method named from() even it is not defining a route and also for every <from> occurrence in a xml file.

davsclaus:
Thanks for the PR

Issue #102, opened 2017-01-05
Restructure source into multi maven module


davsclaus:
We should have a root pom and then maven sub modules for different things.

As we would need a module for a custom camel component which we need for some testing.
Then a module for the idea plugin

And in the future a api module with a reusable api and code for other ide's to reuse etc.

davsclaus:
Will get started on this now but if its too invasive we may do it on a 0.3.x branch and let master be for 0.2.x in case we want to do a 0.2.3 release.

However ideally if this pan out well we should move ahead as it paves the way for supporting custom Camel components.

fharms:
I'm fine with we are doing it now. It's not gonna have a big impact on my task since it mostly new files for the issue I'm working on

davsclaus:
@fharms the code has been merged.

You may need to update your IDE settings a bit, or re-create the Run .. Plugin which I had to do.

fharms:
@davsclaus thanks, I also had to update the location of the plug-in xml file in the module settings under the plug-in tab

PR #110, opened 2017-01-06
#109 Fixed the gutter icon for Java and XML Camel routes

adrianbumbas:


src/main/java/org/apache/camel/idea/util/IdeaUtils.java
```diff
@@ -316,8 +318,13 @@ public static boolean isCamelExpressionOrLanguage(PsiClass clazz) {
...
         PsiMethod method = call.resolveMethod();
         if (method != null) {
-            String name = method.getName();
-            return Arrays.stream(methods).anyMatch(name::equals);
+            PsiClass containingClass = method.getContainingClass();
+            if (containingClass != null) {
+                String className = containingClass.getQualifiedName();
+                String name = method.getName();
+
+                return Arrays.stream(methods).anyMatch(name::equals) && "org.apache.camel.builder.RouteBuilder".equals(className);
```
davsclaus: People can create base classes that extends RouteBuilder, so you would need to look the entire extends hierachy to find that class somewhere. I think PsiClass has some method to get all the extends and their prarents.

davsclaus:
Just a heads up we are restructuring the source code a bit with #102 and a pending PR is in the works.

src/main/java/org/apache/camel/idea/util/IdeaUtils.java
adrianbumbas: I didn't thought about that, now it is fixed. How about the XML route? Checking if <from> is a child of <route> is not enough, <route> must be a child of <camelContext>. What do you think?

PR #114, opened 2017-01-06
support for custom camel components

davsclaus:


camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelService.java
```diff
@@ -100,18 +116,150 @@ public void scanForCamelDependencies(@NotNull Module module) {
...
+                    String[] split = name.split(":");
+                    String groupId = split[1].trim();
+                    String artifactId = split[2].trim();
+
+                    // skip reserved from Apache Camel itself
+                    if ("org.apache.camel".equals(groupId)) {
+                        continue;
+                    }
+
+                    try (URLClassLoader classLoader = newURLClassLoaderForLibrary(library)) {
```
fharms: Would move this out to it's own method to make it more readable

davsclaus: Good idea

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelService.java
```diff
@@ -100,18 +116,150 @@ public void scanForCamelDependencies(@NotNull Module module) {
...
+                                    for (String scheme : part) {
+                                        if (!camelCatalog.findComponentNames().contains(scheme)) {
+                                            // find the class name
+                                            String javaType = extractComponentJavaType(classLoader, scheme);
+                                            if (javaType != null) {
+                                                String json = loadComponentJSonSchema(classLoader, scheme);
+                                                if (json != null) {
+                                                    camelCatalog.addComponent(scheme, javaType, json);
+                                                    // okay a new Camel component was added
+                                                    if (!containsLibrary(artifactId)) {
```
fharms: Wouldn't it make sense to also do this check before all the above code and then continue if it already exists in the library ? It possible you will scan the same library because modules share libraries

fharms:
Damn you are fast @davsclaus :-) 👍

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelService.java
davsclaus: I have checked before with the camel catalog if its a new component it has not seen before, so I only do this once. The ! if check is because a library can have N+ components so I would otherwise add the same artifactId multiple times if a dependency have 2+ components. F.eks. camel-core has 15+ components.

fharms: ok my bad :-)

davsclaus:
Thanks @fharms I just pushed another improvement to make it more clear about that _add_ stuff

PR #120, opened 2017-01-07
8

davsclaus:


camel-idea-plugin/src/main/java/org/apache/camel/idea/CamelProjectComponent.java
```diff
@@ -79,7 +79,8 @@ public void rootsChanged(ModuleRootEvent event) {
                     getCamelIdeaService(project).clearLibraries();
 
                     for (Module module : ModuleManager.getInstance(project).getModules()) {
-                        getCamelIdeaService(project).scanForCamelDependencies(module);
+                        getCamelIdeaService(project).scanForCamelVersionChange(project, module);
```
fharms: maybe we should consider merge them together in one loop to avoid we are looping over the same elements multiple times

davsclaus: We need to separate them because we want to detect the camel version first. Its easier and safter code when its separated.

davsclaus: Yeah @fharms we can optimize this a bit so we only do at most 2 passes
- first is to find camel-core and setup/change the version of the catalog so its all setup
- second to find camel components from Apache Camel itself and 3rd party (and skip already known)

PR #127, opened 2017-01-08
Inspection validate camel endpoints

fharms:
This is the first batch of validation of Camel endpoints and to prevent this will be one big bang feature I decided to deliver in small batches.

This include validation of camel endpoint (properties) and the initial work for validate simple expression.


camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
+/**
```
davsclaus: I would consider renaming this to CamelEndpointAnnotator. We can in the future have other annotators for other kinds, like simple, xpath, data formats, and whatelse there comes down the road.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
...
+ * Validate Camel URI endpoint and simple expression and annotated the specific property
+ * or simple expression to highlight the error on the editor
+ */
+public class CamelAnnotator implements Annotator {
+    @Override
+    public void annotate(@NotNull PsiElement element, @NotNull AnnotationHolder holder) {
+        if (ServiceManager.getService(element.getProject(), CamelService.class).isCamelPresent()) {
+            String fromElement = IdeaUtils.extractTextFromElement(element);
+            if (fromElement != null && !fromElement.isEmpty() && !fromElement.trim().endsWith("&") && !fromElement.trim().endsWith("?")
```
davsclaus: This could be in a static method to return true|false

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
...
+                && !fromElement.trim().endsWith("{{") && !fromElement.trim().endsWith("=")
+                && !fromElement.trim().endsWith("${")) {
+                validateEndpointProperties(element, holder, fromElement);
```
davsclaus: We should do like the smart completion, where we check if its from a known Camel component, eg if the component name is in the list of catalog component names. As otherwise you can trigger on regular string stuff that is not about a Camel endpoint

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }
+    }
+
+    /**
+     * Validate endpoint options list aka properties. eg "timer:trigger?delay=1000&bridgeErrorHandler=true"
+     * if the URI is not valid a error annotation is created and highlight the invalid value.
+     */
+    private void validateEndpointProperties(@NotNull PsiElement element, @NotNull AnnotationHolder holder, String fromElement) {
+        if ((fromElement.contains("?") || fromElement.contains("&")) && IdeaUtils.isConsumerEndpoint(element)) {
```
davsclaus: The Camel endpoints can be anywhere, not only from a consumer endpoint. For example in a .to("jms:cheese?xxx") code line. So I dont think you should limit to only that. Also an Camel endpoint without ? or & can still be validated, as it may be invalid in the context-path section as well. However they may be harder to markup in the code editor where that error is.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
...
+                holder.createErrorAnnotation(range, error);
+            }
+
+        }
+    }
+
+    private void extractMapValue(EndpointValidationResult validateEndpointProperties, Map<String, String> validationMap,
+                                 String fromElement, @NotNull PsiElement element, @NotNull AnnotationHolder holder) {
+        if ((!validateEndpointProperties.isSuccess()) && validationMap != null) {
+            String propertyValue = validationMap.values().iterator().next();
```
davsclaus: The map may contain more keys than only one, for example if you have more options with the same error, eg if you have 2 options that both has an invalid boolean value.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
...
+            return "\tSyntax error: " + result.getSyntaxError();
+        } else if (result.getUnknownComponent() != null) {
+            return "\tUnknown component: " + result.getUnknownComponent();
+        }
+
+        // for each invalid option build a reason message
+        Map<String, String> options = new LinkedHashMap<>();
+        if (result.getUnknown() != null) {
+            for (String name : result.getUnknown()) {
+                if (result.getUnknownSuggestions() != null && result.getUnknownSuggestions().containsKey(name)) {
```
davsclaus: To enable suggestions we need to add camel-catalog-lucene to the dependency, and turn it on at the camel-catalog service. 

See this test: https://github.com/apache/camel/blob/master/platforms/catalog-lucene/src/test/java/org/apache/camel/catalog/lucene/CamelCatalogLuceneTest.java

And just food for thought. Maybe those suggestions can somehow be in IDEA shown as a quick-fix or drowdown list of possible corrections. Just something to work on in the future, if that can be possible.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    // TODO : Solve classloader issue class not found when CamelCatalog access a camel core java file.
+    // The problem is releated to the CamelCatalog is using the plugin classloader, but the Camel core
+    // exits on the project/module classloader
+    /**
+     * Validate simple expression. eg "simple("${body}"
+     * if the expression is not valid a error annotation is created and highlight the invalid value.
+     */
+    private void validateSimpleExpression(@NotNull PsiElement element, @NotNull AnnotationHolder holder, String fromElement) {
```
davsclaus: I would move simple validation out of this class into its own CamelSimpleAnnotator so they are separated.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: I know but I can't mark multiple errors in one annotator

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: I guess we should just run the validator as long we can find a known camel component and not concern about if the URI is still open with &,{{ etc

davsclaus: Yeah you may need to do a bit like this does: https://github.com/davsclaus/camel-idea-plugin/blob/master/camel-idea-plugin/src/main/java/org/apache/camel/idea/documentation/CamelDocumentationProvider.java#L272 however I am not 100% sure if the removal of & or ? is needed, but it may be. Certainly for the XML &amp; you need to unescape it.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: It actually look like you can create multiple annotations with different text range for the same AnnotationHolder. It was not clear in the API but after I look through the code I can see it add a new annotator to a list behind the scene

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: done

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: fixed

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: fixed

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: fixed

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: fixed

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: We know support showing multiple errors

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelAnnotator.java
fharms: Good idea!

fharms:
@davsclaus thanks for the feedback 👍

Issue #94, opened 2017-01-04
Preference page


fharms:
We should allow the user to change some of the preferences for the plug-in through the IDEA preferences dialog

- Add or remove to the property placeholder ignore list
- Change Camel Catalog version
- Show Camel icon in gutter yes|no

This list will properly grow as we extend the plug-in with new features

davsclaus:
The Camel catalog version ought to auto detect what you use, eg 2.18.1 etc. But maybe using the preference you can turn this on / off.

Also if the plugin needs to download a catalog for a different version it has to do this over the internet, which means http proxy issues for some people.

davsclaus:
Also to

- show Camel icon in gutter yes|no

davsclaus:
There is a little example here
https://github.com/forge/intellij-idea-plugin/blob/master/src/main/java/org/jboss/forge/plugin/idea/ui/ForgePreferencePage.java

davsclaus:
So we have an initial preference, then we can add options for these other things.

Its a bit harder to add/remove package names for property key filters as that requires a more advanced UI with some add/remove lines. But we can look at the other plugins and their preferences to be inspired.

davsclaus:
@fharms this PR shows how to add a new option to the preference
https://github.com/davsclaus/camel-idea-plugin/pull/132

PR #134, opened 2017-01-09
#94: allow to turn on|off real time validation

davsclaus:


davsclaus:
This is a preference to turn on or off your new validator. I wonder if the name of the preference and its label is something that the user would understand?

PR #147, opened 2017-01-10
Simple

davsclaus:


camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelEndpointAnnotator.java
```diff
@@ -38,20 +39,41 @@
...
+    void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder, @NotNull String uri) {
+        if (IdeaUtils.isQueryContainingCamelComponent(element.getProject(), uri)) {
+            CamelCatalog catalogService = ServiceManager.getService(element.getProject(), CamelCatalogService.class).get();
 
+            // camel catalog expects &amp; as & when it parses so replace all &amp; as &
+            String camelQuery = uri;
+            camelQuery = camelQuery.replaceAll("&amp;", "&");
+
+            // strip up ending incomplete parameter
+            if (camelQuery.endsWith("&") || camelQuery.endsWith("?")) {
```
fharms: By default I think we should validate what the user see on his screen, otherwise we might end up saying everything is okay but when they run the route it fails

davsclaus: That is a requirement by camel-catalog. If you want to highlight a trailing ? or & then its something you add in the idea plugin.

PR #149, opened 2017-01-10
Val

davsclaus:


davsclaus:
@fharms you are welcome to take a look I forgot to click the review request button.

fharms:
@davsclaus great work..... when I finally get home you have solve all the problems :-)

davsclaus:
No you will have fun with the simple validator to get it on the other languages, and also start using the `getIndex` from the result to better point to where the problem is

However I cannot get an unit test with highlight to mark the warning. It only seems to work for errors. If you turn on warn you get all those compiler errors.

fharms:
yeah it's the issue we have with SDK classes are not resolved when it do it's inspection.

davsclaus:
Ah just found another way. Will do a PR in a short where you can see how I found my warnings.

Issue #91, opened 2017-01-04
Add more unit tests


davsclaus:
Look at adding some more tests using the idea api to test completions, intention etc.

fharms:
Nice work @devclaus 👍😉

davsclaus:
Now its likely you @fharms that is behind on unit tests ;)

fharms:
Damn it :-) 

Well I gonna change that with the Annotator feature

fharms:
Touchdown and the crowd goes wild. With this PR #127 @davsclaus I'm back

davsclaus:
TODO: add unit test for preference

davsclaus:
Not sure its easy to add doc test with smart completion where you simulate pressing ctrl + j, I dont know how to do that.

davsclaus:
Ah found a little way so adding some tests now

fharms:
Closing this since it is not relevant any more. The current level is sufficient and it's mandatory to add new test case for bug fix and new features

PR #150, opened 2017-01-10
#91: added test for validator with warnings about consumer vs produce…

davsclaus:
…r only


fharms:
@davsclaus nice work around until we figure out to setup the SDK as part of the test

PR #154, opened 2017-01-10
Inspection

davsclaus:


camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelEndpointAnnotator.java
```diff
@@ -201,31 +201,7 @@ public String getErrorMessage(EndpointValidationResult result, Map.Entry<String,
...
-                    }
-                }
-            }
-            if (result.getRequired() != null) {
-                for (String name : result.getRequired()) {
-                    returnMsg += "\nMissing required option";
-                }
-            }
-            return returnMsg;
+            return "Unknown option";
```
fharms: We don't want to be smart about which options that are unknown?

fharms: Ahh it's because you can't mark a text range I guess in this case

davsclaus: We dont have the lucene suggestion enabled. And the code was a bit misleading as you get a call per unknown option.

See screenshot where both 2 unknown options are highlighted
screen shot 2017-01-10 at 9 50 23 pm

fharms: But both options are invalid right or what i'm missing here?

davsclaus: Yeah and the code calls that `getErrorMessage` message for each of them, eg first time with `foo` as the name, and then the 2nd time with `dude` as the name.

fharms: I got it. But in case you will have an unknown with suggestions you will not get them

fharms:
@davsclaus awesome work!

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelEndpointAnnotator.java
davsclaus: This was not enabled, see this ticket: https://github.com/davsclaus/camel-idea-plugin/issues/156

PR #159, opened 2017-01-10
Real time simple validator

fharms:
- First version of simple validator for java with getIndex
- Fixes to the one off by one

I push this in small batches because it requires a little more work to test because I have to add the camel-core lib to the module the test framework is creating on setup.

Second it requires a bit more work for supporting other languages because of the different syntax


camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
```diff
@@ -46,7 +46,7 @@ boolean isEnabled() {
      * if the expression is not valid a error annotation is created and highlight the invalid value.
      */
     void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder, @NotNull String text) {
-        boolean hasSimple = text.contains("${") || text.contains("$simple{");
+        boolean hasSimple = text.contains("$simple{");
```
davsclaus: That should be `${` or `$simple{` as it was before. The idea is to only parse/validate the simple text if there is an embedded funktion which must start with `${` or `$simple{`. If there is no embedded functions then its just regular text and no parsing/validation is needed.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelEndpointAnnotator.java
```diff
@@ -91,10 +91,13 @@ private void extractSetValue(EndpointValidationResult result, Set<String> valida
             for (String entry : validationSet) {
                 String propertyValue = entry;
 
-                int propertyIdx = fromElement.indexOf(propertyValue);
+                int startIdxQueryParameters = fromElement.indexOf("?" + propertyValue);
+                startIdxQueryParameters = (startIdxQueryParameters == -1) ? fromElement.indexOf("&" + propertyValue) : fromElement.indexOf("?");
```
davsclaus: I wonder if the endpoint do not have any query parameters, eg without a `?`. Would that not return `-1` and maybe cause the index to be off by one or something?

fharms: this will never be called unless it has ? in the query. It's seems the catalog validation return success = true if the URI has no query parameters

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
fharms: Alright, I have reverted the change

PR #170, opened 2017-01-11
Add support for testing simple expressions

fharms:
More unit test will be added later


davsclaus:
Great work and with some unit tests

PR #174, opened 2017-01-12
simple annotator to validate using predicate or expression and a bit of cleanup

davsclaus:


camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
```diff
@@ -62,43 +75,45 @@ public static boolean isCamelRouteStart(PsiElement element) {
...
         if (IdeaUtils.isFromJavaMethodCall(element, "simple")) {
             return true;
         }
         // xml
-        XmlTag xml = PsiTreeUtil.getParentOfType(element, XmlTag.class);
+        XmlTag xml;
+        if (element instanceof XmlTag) {
+            xml = (XmlTag) element;
+        } else {
+            xml = PsiTreeUtil.getParentOfType(element, XmlTag.class);
```
fharms: Why is this necessary and In which situation do you want to get the parent ?

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
```diff
@@ -126,12 +141,83 @@ public static boolean isCamelRouteSimpleExpression(PsiElement element) {
     }
 
     /**
+     * Is the given element a simple of a Camel route, eg <tt>simple</tt>, ot &lt;simple&gt;.
+     */
+    public static boolean isCameSimpleExpressionUsedAsPredicate(PsiElement element) {
+
+        // java
+        PsiMethodCallExpression call = PsiTreeUtil.getParentOfType(element, PsiMethodCallExpression.class);
```
fharms: Can you just test if if this instance PsiMethodCallExpression?

fharms: I mean this is called when it browse through the PSI tree, right?

fharms:
@davsclaus Looks nice. 👍

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
davsclaus: the element is newer a method call its a literal or something which has a simple text, eg "xxxx" but we need a fair bit of logic to find out which EIP called it, was it filter, transformer etc. Which can be done in different ways in Java DSL and hence all this logic below

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
davsclaus: when the element is a xml attribute for example

PR #176, opened 2017-01-12
Fix quick doc for smart completion list when using prefix multivalue options

davsclaus:


fharms:
@davsclaus looks okay. 

I created a very simple unit test for testing quick lookup of documentation a while back. I think it would be nice if we cover the documentation better with more unit tests

Issue #169, opened 2017-01-11
Realtime Validation : getIndex match the position in the string, but not the position of the caret.


fharms:
Since Camel Catelog have no knowledge about the caret position it report index based on it's location in the String. 

But it could be nice if it was possible to call thecatalogService.validateSimpleExpression with an option to marker the position in the error message so we can replace the position with the editor caret position.

> Unknown function: romuteId at location \<caret\>33\</caret\> Response from Camel CDI on route ${romuteId} using thread: ${threadName}

fharms:
I guess until we get a more secure way to find the location number we can use the "location" in the error string to search for and replace

davsclaus:
The index is the index within the simple string. I dont think it should be the offset from the current caret position in the editor that would be a bit odd imho.

However we could also remove that information and rely on the red underline highlighting, and then its just: `Unknown function: romuteId`.

fharms:
I think is more odd you have to manually count the number of chars in the string rather just using the position from the ide. 

If we don't want to correct the location, then I think it's better we remove the location information as you suggested. 

<img width="804" alt="screen shot 2017-01-12 at 09 27 54" src="https://cloud.githubusercontent.com/assets/1123285/21882176/93698130-d8a9-11e6-9ff6-dcd9e08493e2.png">
<img width="228" alt="screen shot 2017-01-12 at 09 28 11" src="https://cloud.githubusercontent.com/assets/1123285/21882178/95770358-d8a9-11e6-854c-48ac08391bfd.png">

davsclaus:
For example see last image at #172. That tooltip we could just say 'Unknown token *`.

davsclaus:
Yes just having `Unknown function: crouteId` is imho better, then the tooltip is not so long either, and it does not repeat the simple string

fharms:
So we need to parse the error text for either "Unknown token" or "Unknown function" and then strip the rest of the string?

davsclaus:
Yeah I am adding to camel-catalog to provide a shortMessage as well so we can grab that and use.

fharms:
Awesome let me know when it's ready 👍  I'm working on getting support for simple validation for the other languages

davsclaus:
The next camel-catalog build from maven snapshot will have a `getShortError()` method on the result you can use.

fharms:
@davsclaus Btw does this give us a problem with different version of the catalog? I remember something about you can use the same version as camel-core dependency

[...]

PR #177, opened 2017-01-12
#169 Fix the issue with highlight was not displayed correct in some s…

fharms:
- Fix the bug with highlight was not displayed correct in some situation.
- Fix the issue with isCameSimpleExpressionUsedAsPredicate didn't work correctly

We need to add a lot more test cases. It's to easy to break the functionality because the IDEA Annotator and Text range  behave different depended on which file type and type of expression  simple / Predicate


camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
```diff
@@ -69,9 +69,14 @@ void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder,
                     if (!result.isSuccess()) {
                         String error = result.getShortError();
                         TextRange range = element.getTextRange();
+                        int startIdx = result.getIndex() == 0 ? text.indexOf("$") : result.getIndex();
```
davsclaus: Can you add some code comments to this code, its hard to follow what goes on, with the various things. Yes writing code with indexOf and substrings tend to be a bit unclear.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
```diff
@@ -69,9 +69,14 @@ void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder,
...
+
+                        int endIdx = text.indexOf("}", startIdx);
+                        endIdx = endIdx == -1 ? text.indexOf(" ", startIdx) - 1 : endIdx;
+
+                        int propertyLength = (((endIdx < 0 ? text.length() - 1 : endIdx) + 1) - startIdx) + 1;
                         if (result.getIndex() > 0) {
-                            // use -1 to skip the last quote sign
-                            range = TextRange.create(range.getStartOffset() + result.getIndex(), range.getEndOffset() - 1);
+                            range = TextRange.create(range.getStartOffset() + startIdx + 1, range.getStartOffset() + startIdx + propertyLength);
```
davsclaus: We had some stacktraces before when we had this kind of code, where the range would end up having start index > end index. So we should have some safe guard code that checks for this just in case. When it happened before it can happen again.

camel-idea-plugin/src/test/java/org/apache/camel/idea/annotator/CamelSimpleAnnotatorTestIT.java
```diff
@@ -61,14 +61,68 @@ public void testAnnotatorSimpleValidation() {
...
+            + "        }\n"
+            + "    }";
+    }
+
+    private String getJavaWithPredicate() {
+        return "import org.apache.camel.builder.RouteBuilder;\n"
+            + "public class MyRouteBuilder extends RouteBuilder {\n"
+            + "        public void configure() throws Exception {\n"
+            + "              from(\"direct:start\")\n"
+            + "                .loopDoWhile(simple(\"${body.length} <error descr=\"Unknown function: xrouteId\">=!=</error> 12\"))\n"
```
davsclaus: Can you add a 2nd test where you do the variation of this as

    .filter().simple("xxx")

You genererally have 2 ways in java dsl, either as parameter

    .filter ( simple("xxx") )

Or as chained method call

    .filter . simple("xxx")

And that psi parser code is different depending on one other the other are used.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
fharms: We can implement more safeguards and more comment. But the previous solution had the limit that it always assumed it was to the end of the element and to much was highlighted

With the new solution I try to solve 2 issue : 
1. Find the first " " or } after the getIndex
2. if the expression is never ended with " " or "}" we should use the range.getEndOffset()

I will try to construct in another way so it's more readable

camel-idea-plugin/src/test/java/org/apache/camel/idea/annotator/CamelSimpleAnnotatorTestIT.java
fharms: Will do.

fharms:
Close this because there is still on going work

fharms:
@davsclaus let's take another round with this :-)

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
```diff
@@ -168,7 +168,7 @@ public static boolean isCameSimpleExpressionUsedAsPredicate(PsiElement element)
                         element = element.getParent();
                     }
                     if (element instanceof PsiMethodCallExpression) {
-                        exp = (PsiMethodCallExpression) element;
+                        exp = PsiTreeUtil.getParentOfType(element.getParent(), PsiMethodCallExpression.class);
```
davsclaus: this looks a bit wrong, before it was a type cast, you you get the first parent or grand parents that is a method call.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
```diff
@@ -49,7 +50,7 @@ boolean isEnabled() {
      */
     void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder, @NotNull String text) {
         boolean hasSimple = text.contains("${") || text.contains("$simple{");
-        if (hasSimple && CamelIdeaUtils.isCamelSimpleExpression(element)) {
+        if (hasSimple || CamelIdeaUtils.isCamelSimpleExpression(element)) {
```
davsclaus: This should in fact just only be `isCamelSimpleExpression` as that checks if its simple or not. The has simple was to test for simple functions, but you can do simple without functions, such as `5 < 10'. But not so sure if they make so much sense.

davsclaus:
@fharms there is a merge conflict now due the travis file has been merged

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
fharms: previous it didn't go far enough back in the PSI tree to find parent method and therefore it didn't see the simple method was an input predicate to another method.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
fharms: I previously removed this and you made comment, that it would not work for expression like this
`.log("Result from query ${body}")`

Is the CamelIdeaUtils.isCamelSimpleExpression(element) handling this case also?

fharms: If I remove the 

> boolean hasSimple = text.contains("${") || text.contains("$simple{");

the test will failed for expressions like this
> .log("Result from query ${body}")

davsclaus: Yes but its still wrong. We need to add special support for log. Its only for simple / <simple> currently.

Otherwise you end up validating all kind of non Camel stuff as well.

My PR will correct this and have a TODO for log

fharms:
@davsclaus see my comments. I might be wrong or my test case is wrong, but previous it didn't work with code like this:

```java
 from("direct:start")
                .loopDoWhile(simple("body.length !s1= 12"))
                .to("mock:loop")
                .transform(body().append("A"))
                .end()
                .to("mock:result");
```

davsclaus:
Camel simple is only for `${ }` functions not without that. Its deprecated without functions and the old parser is being removed from Camel.

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
fharms: So if do something like this in CamelIdeaUtils.isCamelSimpleExpression(element) we should be safe?

>  if (IdeaUtils.isFromJavaMethodCall(element, "simple", "log")) {
            return true;
        }

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
fharms: Okay, revert my change here

PR #184, opened 2017-01-13
Fixes #155 to only accept xml tags from known namespaces

### LINKED ISSUE #155: Detecting Camel endpoint in xml files may pick non Camel
We can sometime hit mistakes like a <bean> which refers to a AMQ broker which can use its own internal vm protocol to talk to it inside the same jvm.

screen shot 2017-01-10 at 9 45 48 pm

davsclaus:


fharms:
Look good

Issue #157, opened 2017-01-10
Camel facet


davsclaus:
Spring has a facet so when IDEA have detected its a spring project with spring xml file or spring boot etc, then it can setup a facet for that.

We could have a Camel facet, which has its own little Camel window, which can have a list of all the CamelContext / routes there is in the project, so you can navigate among them.

adrianbumbas:
Things done so far:

- Enabled Camel facet

- The facet configuration page is empty

- Facet autodetection works only for XML files

What is difficult is to make it work for any other file types, apparently each file type needs another facet.

adrianbumbas:
Apparently we need a facet for every file type 
https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000031990-Facet-detector-for-multiple-file-types

davsclaus:
I dont think a facet etc brings so much value. We can open tickets in the future if we find something cool to add

PR #186, opened 2017-01-13
#157 Added first version of facet plus autodetection for XML files

adrianbumbas:


camel-idea-plugin/src/main/java/org/apache/camel/idea/facet/CamelFrameworkDetector.java
```diff
@@ -0,0 +1,60 @@
...
+    @NotNull
+    @Override
+    public FacetType<CamelFacet, CamelFacetConfiguration> getFacetType() {
+        return CamelFacetType.getInstance();
+    }
+
+    @NotNull
+    @Override
+    public FileType getFileType() {
+        return StdFileTypes.XML;
```
davsclaus: I wonder if we can have ANY or something as file tile?

camel-idea-plugin/src/main/java/org/apache/camel/idea/facet/CamelFrameworkDetector.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @NotNull
+    @Override
+    public ElementPattern<FileContent> createSuitableFilePattern() {
+        return FileContentPattern.fileContent().with(new PatternCondition<FileContent>("") {
+            @Override
+            public boolean accepts(@NotNull final FileContent fileContent, final ProcessingContext context) {
```
davsclaus: And here we can check if the file is xml, java, kotlin, groovy or scala which is the files we support. And then also check if there is Camel inside the file, as you do.

camel-idea-plugin/src/main/java/org/apache/camel/idea/facet/CamelFacetConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class CamelFacetConfiguration implements FacetConfiguration {
+    @Override
+    public FacetEditorTab[] createEditorTabs(FacetEditorContext facetEditorContext, FacetValidatorsManager facetValidatorsManager) {
+        return new FacetEditorTab[]{new FacetEditorTab() {
+            @NotNull
+            @Override
+            public JComponent createComponent() {
+                return new JLabel("Content goes here");
```
davsclaus: I guess here is the UI for the facet where we can list the Camel routes discovered. And if the user clicks a route, it opens the file and position the cursor where the route starts.

adrianbumbas: Yes, this is the place for the facet UI. A facet is not meant to be used to navigate, it is for configuration only (see https://www.jetbrains.com/help/idea/2016.2/facet.html).
For navigation i think the best is to use a tool window like Spring provides.

camel-idea-plugin/src/main/java/org/apache/camel/idea/facet/CamelFrameworkDetector.java
adrianbumbas: There is nothing similar to ANY.

adrianbumbas: I've ask the question on the Intellij community forum https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000031990-Facet-detector-for-multiple-file-types

camel-idea-plugin/src/main/java/org/apache/camel/idea/facet/CamelFacetConfiguration.java
adrianbumbas: @davsclaus What would be the purpose of the facet? Configuration, navigation to the routes or both? If is for navigation only maybe we should look to other components and leave the facet aside.

davsclaus: @adrianbumbas yeah lets research a bit what IDEA offers and what is best. I wanted to have a window etc that scans your project and detect all your Camel routes, and list them, so you can see them in one window, and then click on the route to open the file in the source editor and position the cursor on the line where the route starts.

I guess facet is not the answer. If facet is better for project configuration then for Camel that makes less sense. If you use Camel with Spring Boot, the the SB facet is for configuring SB and Camel can do that too with SB configuration.

adrianbumbas: @davsclaus ok, let's put the facet aside. I think that the facet should exists at least for autodetection even there is no configuration, it gives a good feedback to the user. I will close this PR for now.

PR #188, opened 2017-01-14
Add preference to switch to use custom icon

davsclaus:
down the road we will ship with a set of embedded icons so users can easily switch between them. But for now lets allow graphic arts to try different icons easily by loading from their file system


camel-idea-plugin/pom.xml
```diff
@@ -116,6 +116,13 @@
       <version>${snakeyaml.version}</version>
     </dependency>
 
+    <!-- Swing MigLayout -->
+    <dependency>
+      <groupId>com.miglayout</groupId>
+      <artifactId>miglayout-swing</artifactId>
+      <version>5.0</version>
```
fharms: move the version to a property like we do with the other dependencies

PR #190, opened 2017-01-14
187

davsclaus:


camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelSimpleAnnotator.java
```diff
@@ -49,7 +49,8 @@ boolean isEnabled() {
      */
     void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder, @NotNull String text) {
         boolean hasSimple = text.contains("${") || text.contains("$simple{");
-        if (hasSimple && CamelIdeaUtils.isCamelSimpleExpression(element)) {
+        if (hasSimple
+            && CamelIdeaUtils.isCamelSimpleExpression(element)) {
```
fharms: This should be OR. Otherwise it would not work with 
if (hasSimple
             || CamelIdeaUtils.isCamelSimpleExpression(element)) {

    .filter().simple("xxx")

You genererally have 2 ways in java dsl, either as parameter

    .filter ( simple("xxx") )

Or as chained method call

    .filter . simple("xxx")

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/IdeaUtils.java
```diff
@@ -428,6 +428,91 @@ public static boolean isFromGroovyMethod(PsiElement element, String... methods)
...
+
+        if (found) {
+            kind = element.toString();
+            // must be an identifier which is part of the method call
+            if (kind.contains("identifier")) {
+                String name = element.getText();
+                boolean match = Arrays.stream(methods).anyMatch(name::equals);
+                if (match) {
+                    // sanity check that "from" was from a from a method call
+                    PsiElement parent = element.getParent();
```
fharms: If "think" you can use the line below to find the parent method and test for (exp instanceof PsiMethodCallExpression) instead.

exp = PsiTreeUtil.getParentOfType(element.getParent(), PsiMethodCallExpression.class);

PR #192, opened 2017-01-14
Enable Travis CI Cache feature

fharms:
- Enable Travis CI Cache feature
- Test if Intellij artifacts is already installed.

PR #194, opened 2017-01-15
Fix shortname override

fharms:
The Javadoc for getShortName specific says you should not overwrite this method
and the result is an ERROR is throw


fharms:
@davsclaus this PR should first be merged after the PR #192 merged

PR #198, opened 2017-01-15
groovy simple annotator for real time validation

davsclaus:


davsclaus:
Yeah down the road that would be great, but you can do a bit of manual testing by loading the camel-example-groovy and adjust the code a bit to see how it works,

PR #199, opened 2017-01-15
191

davsclaus:


davsclaus:
Ah yeah need to wait for new camel snapshot build to have the API change we needed
https://builds.apache.org/view/A-D/view/Camel/job/Camel.trunk.notest/2996/

Issue #197, opened 2017-01-15
Simple validation - Add support for log EIP


davsclaus:
The log EIP uses simple expression in the log message.

    log("Hello ${body}")

   And in XML

    <log message="Hello ${body}"/>

We need to add special support for this in the simple annotator.

davsclaus:
@fharms so this requires to add some new code that checks its the "log" method call or `<log>` XML tag.

davsclaus:
And there is an unit test with `@Ignore` that can be enabled when this has been implemented

PR #200, opened 2017-01-15
Simple validation - Add support for log EIP #197

fharms:


camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
```diff
@@ -111,7 +111,7 @@ public static boolean isCamelRouteStart(PsiElement element) {
      */
     public static boolean isCamelSimpleExpression(PsiElement element) {
```
davsclaus: Can you add to the javadoc that we also check the log EIP as well

fharms:
@devsclaus we have reached pr nr 200. Let's celebrate this afternoon with a beer😊🍺

davsclaus:
Yep was just told the neighbors will come by for a beer. We have some x-mas beers left.

Issue #221, opened 2017-01-17
Add icons for retrina displays


davsclaus:
See
http://www.jetbrains.org/intellij/sdk/docs/reference_guide/work_with_icons_and_images.html

We need to have a 32x32 icon size with the name `camel@2.png` and `camel2@2.png`.

Anyone wanna help resize those icons.

The original is here
https://github.com/apache/camel/pull/1369

And another is here
https://github.com/fabric8io/fabric8-console/blob/master/img/icons/camel.svg

PR #227, opened 2017-01-20
Added 32x32 icons

MarkoMackic:


MarkoMackic:
Referencing issue #221

davsclaus:
Thanks. Its the PNG files that is what is actually being used by the plugin. The svg are _just_ the original source of the icon. I wonder if you could try to convert the svg to 32x32 png files?

I have found online converters that can do this on the internet in lack of a good app.

MarkoMackic:
No problems, I'll commit it today

davsclaus:
Thanks a lot

Issue #228, opened 2017-01-20
Quick documentation popup incorrect title for XML


adrianbumbas:
The quick doc popup doesn't show the correct element in the title for which doc is displayed. This is happening in routes where & is present or when the caret is at the end of the route.

Examples:
Caret is at the end of route definition
![route_end](https://cloud.githubusercontent.com/assets/1636509/22164930/acff5e56-df59-11e6-96e4-c21ecf6560fb.png)
Caret is on the &amp;
![caret_amp](https://cloud.githubusercontent.com/assets/1636509/22164929/acfe0d76-df59-11e6-9848-f70a76ae7441.png)

davsclaus:
Thanks for spotting and fixing this

PR #229, opened 2017-01-20
#228 Fixed quick doc title

adrianbumbas:
Fix for #228


fharms:
It would be preferable we have this unit tested also

davsclaus:
I think a test will be hard as the doc was also shown before, it was only the title in the dialog that was not showing correctly. Dont think you can test that from unit test.

adrianbumbas:
I will try to add a test anyway. I'm having issues detecting the PsiElement at the caret position.

PR #250, opened 2017-02-01
Fix smart completion

fharms:
This should fix the known problems with smart completion for options, enums and values.  

One of the challenges with smart completion is the different possibilities of various cases it should work with and to cover for this I have added additional test cases both for Java and Xml

We still need to figure out how we can test if the user have marked a word and press tab instead of enter when selecting the word.


fharms:
@davsclaus it was and I hope I cover most cases in my test. If an issue pops up we should create a test case for it

davsclaus:
I got 4 checkstyle errors compiling latest code. Maybe you can do a PR to fix those if you have time tonight. If not then its surely fine and I can fix that too when I hack on the code later this evening

fharms:
I will create a PR to fix them

Issue #242, opened 2017-01-27
Camel route start is incorrectly detected for Java 


adrianbumbas:
The method `CamelIdeaUtils.isCamelRouteStart` returns true for all the `PsiElements` of a Java Camel route.

Example: for `from("file:inbox")` it returns true for the following elements:

<img width="512" alt="screen shot 2017-01-27 at 21 34 43" src="https://cloud.githubusercontent.com/assets/1636509/22386853/40d3dd12-e4d9-11e6-9acd-debf9bffa1f5.png">

It should return true only for "file:inbox" just like for XML.

adrianbumbas:
Added unit test to prove that in PR #243

adrianbumbas:
I was able to reduce the number of detected elements from 8 to 2 :)

To continue further is very tricky because `"file:inbox"` is at the same time a `PsiJavaToken` and `PsiLiteralExpression` but a `PsiJavaToken` is not a `PsiLiteralExpression`. The annotator needs a `PsiLiteralExpression` and the completion needs a `PsiJavaToken`.

Any suggestions?

davsclaus:
Maybe you can add a parameter that includes which psi element you want to use or something. Also mind that in XML then the psi elements are different, and so are they in groovy, scala etc.

adrianbumbas:
I'm trying to use that no success so far, the annotator test are failing anyway, not sure why.
Relying on string literals is not the best option since this route

` String uri = "direct:cafe";`
 `from(uri)...`

is not detected as being a Camel route.

davsclaus:
Ah yeah the latter I dont think is yet supported as it requires a bit more logic to determine this and find out its a reference to a variable, and where that variable is, and find out if that is a string etc, or a `org.apache.camel.Endpoint` type etc.

davsclaus:
We should fix so there is only one Camel icon shown, eg in the line where `<from>` or `from` is. See for example those 4 icons at this picture: https://cloud.githubusercontent.com/assets/1123285/22860595/f177158c-f102-11e6-8423-b5838be42e81.gif

davsclaus:
This is fixed recently

adrianbumbas:
@davsclaus why do you think that this is fixed, the unit test still shows that there are 8 routes:

[...]

Issue #240, opened 2017-01-24
Camel route navigator


adrianbumbas:
When the user clicks on the Came gutter icon, a list of related routes will appear that allows to users to navigate to them.

The proof of concept presented in #108 uses a brute force approach: it scans all the files from the project looking for PsiElements which is very slow.

I'm trying to leverage the Search API which is much faster, the only issue is that it supports search by word, not strings:

https://intellij-support.jetbrains.com/hc/en-us/community/posts/206770865-Why-is-findFilesWithPlainTextWords-not-working-

adrianbumbas:
I was able to use the Search API and I have a working version that navigates from Java
 to XML routes. The only issue is that there are duplicate navigation entries for Java endpoints, i'm blocked by #242 

This is how it looks so far, i'm going to customize a little bit this popup to remove the full path to file and to add a Camel icon.
<img width="1403" alt="screen shot 2017-01-29 at 13 53 06" src="https://cloud.githubusercontent.com/assets/1636509/22404110/610c83b6-e62a-11e6-89aa-89d740e12db7.png">

fharms:
Closing this because it's already merged

PR #258, opened 2017-02-05
#240 Added gutter route navigation from/to Java/XML

adrianbumbas:
First version of the navigation that suports Java literals and XML only. The issue #242 is still not solved so i did a workaround to filter out everything that is not a literal for Java.

### DIFF
--- a/camel-idea-plugin/src/main/java/org/apache/camel/idea/gutter/CamelRouteLineMarkerProvider.java
+++ b/camel-idea-plugin/src/main/java/org/apache/camel/idea/gutter/CamelRouteLineMarkerProvider.java
@@ -16,15 +16,32 @@
  */
 package org.apache.camel.idea.gutter;
 
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
+import java.util.List;
 import javax.swing.*;
 
 import com.intellij.codeInsight.daemon.RelatedItemLineMarkerInfo;
 import com.intellij.codeInsight.daemon.RelatedItemLineMarkerProvider;
 import com.intellij.codeInsight.navigation.NavigationGutterIconBuilder;
+import com.intellij.lang.java.JavaLanguage;
 import com.intellij.openapi.components.ServiceManager;
+import com.intellij.openapi.editor.markup.GutterIconRenderer;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
+import com.intellij.openapi.project.Project;
 import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiLiteralExpression;
+import com.intellij.psi.PsiMethodCallExpression;
+import com.intellij.psi.impl.source.tree.java.PsiLiteralExpressionImpl;
+import com.intellij.psi.impl.source.xml.XmlTagImpl;
+import com.intellij.psi.search.GlobalSearchScope;
+import com.intellij.psi.search.PsiSearchHelper;
+import com.intellij.psi.search.UsageSearchContext;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.intellij.psi.xml.XmlElementType;
+import com.intellij.psi.xml.XmlTag;
+import com.intellij.psi.xml.XmlToken;
 import org.apache.camel.idea.service.CamelPreferenceService;
 import org.apache.camel.idea.service.CamelService;
 import org.apache.camel.idea.util.CamelIdeaUtils;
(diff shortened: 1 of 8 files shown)


adrianbumbas:
@davsclaus feel free to merge this if you don't have any comments

camel-idea-plugin/src/main/java/org/apache/camel/idea/gutter/CamelRouteLineMarkerProvider.java
```diff
@@ -69,4 +93,116 @@ private CamelPreferenceService getCamelPreferenceService() {
...
+     *
+     * @param route      the complete Camel route to search for
+     * @param psiElement the {@link PsiLiteralExpression} that might contain the complete route definition
+     * @return the {@link PsiElement} that contains the exact match of the Camel route, null if there is no exact match
+     */
+    private PsiElement findJavaElement(String route, PsiLiteralExpression psiElement) {
+        if (route.equals(psiElement.getValue())) {
+            //the method 'to' is a PsiIdentifier not a PsiMethodCallExpression because it's part of method invocation chain
+            PsiMethodCallExpression methodCall = PsiTreeUtil.getParentOfType(psiElement, PsiMethodCallExpression.class);
+            if (methodCall != null && "to".equals(methodCall.getMethodExpression().getReferenceName())) {
```
davsclaus: You can also have `toD` or `toF` in Camel. The former is used more than the latter which is a shortcut for using String.format. to build the uri string.

camel-idea-plugin/src/main/java/org/apache/camel/idea/gutter/CamelRouteLineMarkerProvider.java
```diff
@@ -69,4 +93,116 @@ private CamelPreferenceService getCamelPreferenceService() {
...
+    /**
+     * Further refine search in order to match the exact XML Camel route.
+     *
+     * @param route      the complete Camel route to search for
+     * @param psiElement the {@link PsiElement} that might contain the complete route definition
+     * @return the {@link PsiElement} that contains the exact match of the Camel route
+     */
+    private PsiElement findXMLElement(String route, XmlToken psiElement) {
+        if (psiElement.getTokenType() == XmlElementType.XML_ATTRIBUTE_VALUE_TOKEN) {
+            if ("to".equals(PsiTreeUtil.getParentOfType(psiElement, XmlTag.class).getLocalName())) {
```
davsclaus: In XML DSL there is only `<to>` and <toD>`

camel-idea-plugin/src/main/java/org/apache/camel/idea/gutter/CamelRouteLineMarkerProvider.java
```diff
@@ -69,4 +93,116 @@ private CamelPreferenceService getCamelPreferenceService() {
...
+     * Defines the search scope for the Camel routes.
+     * It consist of all the files from all the modules including the files from JAR files.
+     *
+     * @param project the current {@link Project}
+     * @return the search scope
+     */
+    private GlobalSearchScope getSearchScope(Project project) {
+        Module[] modules = ModuleManager.getInstance(project).getModules();
+
+        GlobalSearchScope searchScope = GlobalSearchScope.EMPTY_SCOPE;
```
davsclaus: Are we not searching to far? I think you should only search in the project source code, eg `src/main/java` and maybe also `src/test/java` etc. Not in JARs and other libraries.

camel-idea-plugin/src/test/resources/testData/JavaCamelRouteLineMarkerProviderTestData.java
```diff
@@ -1,38 +1,29 @@
...
 
-public final class JavaCamelRouteLineMarkerProviderTestData {
-
-    private JavaAddEndpointIntentionTestData() { }
-
-    public static void main(String[] args) {
-        Main main = new Main();
-        main.addRouteBuilder(new MyRouteBuilder());
-        main.run(args);
+public class JavaCamelRouteLineMarkerProviderTestData extends RouteBuilder {
```
davsclaus: I think this PR has checkstyle errors, when you build the code it should report errors.

davsclaus:
@adrianbumbas lets merge this and then you can do another PR with the minor improvements I pointed out. 

It would be good to get a release out with latest stuff, and then refine and get community feedback

adrianbumbas:
@davsclaus thanks, i will do the requested changes.

Issue #264, opened 2017-02-08
Route navigation - refine search scope


adrianbumbas:
In order to find the Camel routes currently a search is made in the whole project including in files that cannot contain Camel routes like *.xsd, *.html, pom.xml etc

@davsclaus it makes sense to look also in the jar files? What if the routes are defined there?

davsclaus:
IMHO it does not make sense to search in JAR files as users tend to have all their Camel routes in the same project. 

For files to search for routes then we have some code that can check if its from supported file extensions eg java, xml, scala etc. Just search the code a bit and you find that.

adrianbumbas:
@davsclaus can we close this?

PR #271, opened 2017-02-12
#264 Refined Camel route search scope

adrianbumbas:


adrianbumbas:
For some reason this test fails from Maven but works when being run from IntelliJ. Does somebody have any idea why?

davsclaus:
I am not sure why it fails

fharms:
@adrianbumbas there is a few things you need to aware of when building from maven. 

1. It could be a dependency problem because of missing jar file.
2.The configuration in the failsafe maven plugin is customized for running the plugin test from maven. It's possible you need to configure additional properties.

My advice is to look through the log is printed out when running the build, and you might find the smoking gun. That at least what I have do in the past

adrianbumbas:
I think that the problem is at the `maven-failsafe-plugin` configuration, it seems that the JVM options defined in the ` <argLine>` section are not passed to the test.

What is strange is that the `CamelProjectComponentTestIT` needs also the correct installation path but this one works. If I change `CamelRouteSearchScopeTest` to extend `ModuleTestCase` like `CamelProjectComponentTestIT` it doesn't work!

PR #273, opened 2017-02-13
Multiline completion support

fharms:
This is the first batch of multi line support for code completion, if you split endpoint uri over multiple lines the code completion will still work
```java
  from("timer:trigger?repeatCount=10" +
                "&fixedRate=false&bridgeErrorHandler=false" +
                "&daemon=false" +
                "&period=10&delay=1000")
```

The validation is not tested in depth, but a quick test shows it worked in the cases I tried. 
Since this required changes in some of the shared methods I want to push this to avoid too many conflicts. The code is in a good shape and multiple tests cases is added, but after it's merged I will properly use some time to clean it up a bit and test validation more


davsclaus:
Nice work

Issue #265, opened 2017-02-08
Route navigation - do not search only for "to" routes


adrianbumbas:
Look also for routes defined in Java files in the methods `toD` and `toF` and for XML in `toD`

adrianbumbas:
@davsclaus can we close this?

davsclaus:
Yeah sure you are welcome to close issues yourself as well

PR #274, opened 2017-02-14
#265 Detect Camel routes in Java methods toD() and toF() and in XML i…

adrianbumbas:
…n toD elements


adrianbumbas:
@davsclaus I see that the unit tests are failing after you did the merge, this didn't happened when I did the pull request so I guess that the order in the list is not guaranteed to be the same. I will check this issue.

PR #276, opened 2017-02-16
added support for sbt

rparree:
I added support for SBT. I tested it and it works. 

p.s. I noticed however for scala routes the gutter icon does not appear (completion on uri parameters works). I'll have a look why the gutter icon does not appear.


davsclaus:
Also beware of changing the order of the java imports, as the build uses a checkstyle check where it reports errors if the code formatting is not correct.

rparree:
This can be closed as you already added support for sbt. I leave it up to you to close it.

davsclaus:
Ah thanks closing

Issue #277, opened 2017-02-16
Scala routes using "==>" have no gutter


rparree:
There is no gutter for scala routes.

Example route:

```scala
import org.apache.camel.CamelContext
import org.apache.camel.scala.dsl.builder.{RouteBuilder, ScalaRouteBuilder}


class ScalaCamelRouteLineMarkerProviderTestData(ctx : CamelContext ) extends ScalaRouteBuilder(ctx) {
    "file:inbox" ==> {
        -->("file:outbox")
    }
}
```

I am experimenting with [org.apache.camel.idea.util.CamelIdeaUtils#isCamelRouteStart](https://github.com/rparree/camel-idea-plugin/blob/e424c3bf92189865a52003a1e653ba83b189505e/camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java#L105). I am however not experienced enough yet with PSI trees. It feels like the code should look for an infix epxession* something along the lines of 

```java
if (element instanceof LeafPsiElement) {
    IElementType type = ((LeafPsiElement) element).getElementType();
    if (type.getLanguage().isKindOf("Scala")) {
        try {
            // TODO needs cleaning and remove try...catch (plus does not work)
            PsiElement infixExpression = element.getParent().getParent();
            return infixExpression.getChildren()[1].getText().equals("==>");
        } catch (NullPointerException | ArrayIndexOutOfBoundsException ignored){

        }
    }
}
```

But that is actually giving me too many gutter icons

(Non working) code at https://github.com/rparree/camel-idea-plugin/tree/scala-gutter

PR #278, opened 2017-02-16
fix for #277: Scala routes using "==>" have no gutter icon

rparree:
I have tested it with my scala dsl demos and so far so good. 

It does not support routes like using an implicit conversion without the arrow method. Most likely there is a way to obtain implicit in scope to see if the string gets converted using the implicit def `stringToRoute`. I'll can a look into that later.


davsclaus:
Hi thanks for contributing. As this PR now have merge conflicts do you mind updating this?

davsclaus:
We have merged the other PR so closing this

Issue #187, opened 2017-01-14
Simple annotator - Add support for scala, groovy and kotlin


davsclaus:
These need a bit more work to find out simple -> EIP which called them, and if its a predicate vs expression.

davsclaus:
Groovy is now implemented.

davsclaus:
A bit more complex scala code and it gets much more complex to find out this with all this AST parent / child / sibling check.

rparree:
I'll have a look at this. Is the `org.apache.camel.idea.util.IdeaUtils#isPrevSiblingFromScalaMethod` method i am guessing that needs some work?

davsclaus:
Yeah

davsclaus:
Supporting those extra languages like Groovy and Scala is IMHO becoming a burden. 

So lets try to leave it as-is. Or find some way to make it optional and move to another maven module or something.

PR #280, opened 2017-02-17
fixed #277: Scala routes using "==>" have no gutter icon

### LINKED ISSUE #277: Scala routes using "==>" have no gutter

rparree:
created a new pull request

### DIFF
--- a/camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
+++ b/camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
@@ -15,21 +15,31 @@
  * limitations under the License.
  */
 package org.apache.camel.idea.util;
-
+import java.util.Arrays;
+import java.util.Optional;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
-import com.intellij.psi.*;
+import com.intellij.psi.PsiAnnotation;
+import com.intellij.psi.PsiClass;
+import com.intellij.psi.PsiClassType;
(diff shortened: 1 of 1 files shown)


rparree:
I'll have a look at the checkstyle error

davsclaus:
@rparree as a Scala user you are welcome to look at #187 as we need the plugin to support validating camel simple expression in scala dsl.

PR #286, opened 2017-02-27
Ant filepath pattern

fharms:


camel-idea-plugin/src/main/java/org/apache/camel/idea/service/CamelPreferenceService.java
```diff
@@ -57,6 +57,10 @@
         // ignore camel-spring-boot auto configuration prefixes
         "camel.springboot.", "camel.component.", "camel.dataformat.", "camel.language."};
 
+    @Transient
+    private static String[] defaultExcludeFilePattern = {
+        "**/**/log4j.properties", "**/**/logging.properties"};
```
davsclaus: I wonder if `**/log4j.properties' is not sufficient. Do we really need that 2 x ** ?. Also there is log4j v2, so try adding "log4j2.properties" as well. For example we use that in Camel: https://github.com/apache/camel/tree/master/camel-core/src/test/resources

fharms: Good point

camel-idea-plugin/src/main/java/org/apache/camel/idea/preference/CamelExcludePropertyFileTable.java
```diff
@@ -174,7 +174,7 @@ public void add(@NotNull final String property) {
 
         @Override
         public void addRow() {
-            String pattern = Messages.showInputDialog("", "Enter filename pattern", null);
+            String pattern = Messages.showInputDialog("", "Enter filename pattern (**/**/file.properties)", null);
```
davsclaus: Can you update this example as well

PR #298, opened 2017-04-08
Update links to match the new repository and upgraded intellij to the…

fharms:
… latest version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ cache:
     - $HOME/.m2
 before_install:
     -  "mkdir idea"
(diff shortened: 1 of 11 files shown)


fharms:
Of course upgrading Intellij is not smoothless process. *sign*

Problem with running the unit test after I fixed the maven dependency problem

davsclaus:
Ah darn could a temporary workaround be to keep the 2016.x version and then do the 2017.x upgrade in a separate PR?

fharms:
unfortunately Intellij prevent download of older versions, so it will not even work if I revert the upgrade. Only solution is to move forward

PR #322, opened 2017-10-07
Add tests for Stringutils

bennetelli:


fharms:
It looks like we have to upgrade to IDEA 2017.2.5. 

I can fix the upgrade, but that mean you will have to download the artifacts for 2017.2.5 after like explained in the README.md

fharms:
If you pull the latest and run : 

./download-unpack-intellij.sh 2017.2.5 idea idea-IC-172.4343.14 $HOME/.m2
./install-intellij-libs.sh 2017.2.5 idea $HOME/.m2  

You should be fine

bennetelli:
this should fix my errors when running the tests, right? Wanted to do something else in the meantime and don't spend so much time in fixing the errors reported in my issue. Therefore I decided to write some additional plain JUnit Tests which are not affected by my issue.

bennetelli:
don't understand the travisCI check. Does it have anything to do with my changes in this PR? :/

davsclaus:
@fharms wonder if we have a comment we can type here to tell circle ci to re-test this

fharms:
@davsclaus I'm not sure I can look into, but in this case it didn't matter because the pom.xml and travis .xml was point to an old version of IDEA

davsclaus:
Yeah it looks fine to merge, we can try to close and reopen however

davsclaus:
trying to kick off ci build

bennetelli:
is the checkstyle configuration on TravisCI different than the one in the project?

PR #326, opened 2017-10-07
Remove cyclic dependencies between packages

bennetelli:
we should keep an eye of dependencies between packages to avoid an unmaintainable-big-ball-of-mud-plugin. So I used structure101 to analyze dependencies between packages/classes and moved some methods to another location.
Please let me know if it doesn't make sense to move these methods to these classes.


davsclaus:
Thanks

PR #334, opened 2017-10-12
Update checkstyle and add unused import check

bennetelli:
just saw that there are some unused imports and we don't have a checkstyle check for it because there was a bug in the checkstyle version we used. So I updated checkstyle to a version where this issue is fixed (see: https://github.com/checkstyle/checkstyle/issues/2840)


bennetelli:
btw: why do we want to meet the ActiveMQ coding guidelines? 

There is a comment in the camel-checkstyle.xml: "Checks to make sure the code meets the ActiveMQ coding guidelines..."

davsclaus:
@bennetelli yeah that was a copy, so you can change the name to Apache Camel instead of AMQ

bennetelli:
@davsclaus and what about the source camel-checkstyle.xml in https://github.com/apache/camel/tree/master/buildingtools ?

davsclaus:
Yeah sure you are welcome to provide a PR for that

PR #335, opened 2017-10-12
Fix some code issues

bennetelli:


bennetelli:
it's better to squash the commits isn't it?

fharms:
@bennetelli personally I like the idea as long the commit msg include headlines for all the various changes and they all belong to the same "change", but it's not something we really have talked about yet.

PR #337, opened 2017-10-13
allow arrow up and down to switch lines in property tables

bennetelli:
currently clicking up and down arrow leads to an edit-popup-dialog. This commit fixes this issue


bennetelli:
Nevertheless there is another error left and I'm currently rewriting the page because I think it should be possible to get the same layout with less code

PR #350, opened 2017-10-22
Support optional plugin support

fharms:
This is the initial pull-request for supporting optional plugin support using extension. With the depreciation of Scala, groovy and kotlin support we really don't have any optional dependencies, but the work with extract the language specific code to extension point still make a lot of sense for Java, Xml, Properties and yaml.

It only take 2 methods from the IdeaUtils and refactor them to use the extension point instead, because I want to take this in steps and it also make possible for others to help on this.

The idea was to cut the API at method level for each utility so we preserve the current API for CamelIdeaUtils, IdeaUtils etc but let each method call it’s extension points. This has the advancement of not changing the API against all the consumers using the CamelIdeaUtil, IdeaUtils etc utils, and the current Utils will just serve as a “wrapper”


fharms:
The extension systems seems pretty simple, but I guess we could let the extensions extend a base class where we can do some of the magic to determine what type of language. 

But the end of the day we will have to loop through the extensions to figure out if Java,XML etc, but we can possible do it in a smarter way.

I will give it some additional thoughts

bennetelli:
@fharms looks good so far :) btw: what is that order of extensions in plugin.xml good for?

PR #373, opened 2018-01-29
update plugin preferences documentation and add ultimate stuff

bennetelli:
* update plugin preferences so they contain the current preferences
* add missing dependency for IDEA Ultimate 
* add idea.version tag for Maven ultimate profile


bennetelli:
oops. looks like there where some additional commits in my commit log. @davsclaus let me know if it's a problem or if I can just squash and merge both changes.

davsclaus:
Oh if you can merge it then just go ahead.

PR #379, opened 2018-02-03
Update to idea 2017 3 4

bennetelli:


bennetelli:
updated my IDEA Ultimate to 2017.3.4 and it's working. I also checked the buildnumber of CE and Ultimate and they are similar. In my opinion this issue would be fixed, but it would be nice if you @fharms could take a look at it

PR #384, opened 2018-02-16
fix erroneous html list item

bennetelli:


bennetelli:
@davsclaus I also updated this tag:
 <idea-version since-build="173.0"/>

it pointed to idea build 162 before. I'm wondering if this tag is still relevant? Is it necessary to update it everytime a new IDEA version is released?

davsclaus:
No that build number is essentially the oldest version of IDEA the plugin can run on.

PR #385, opened 2018-02-16
Camel java dsl bean method completion

fharms:
This is the first version of the feature for method smart completion
from Java Camel DSL routes. The first version has limit support for
method completion and following functionality is implemented.
- Working with '.bean(MyClass.class, "<caret>"'
- Show method public methods from base class and it's super classes
- Mark @Handle method as blod in the completion list
- Mark deprecated as String through
- Right side of the completion list is the containing class
- Jump to declaration is only supported from the Camel Java DSL
- Rename of method name in Camel Java DSL and bean is supported
- Test case for testing bean method completion Java bean class hierarchy
- Test case for testing rename.


camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
```diff
@@ -180,6 +171,35 @@ public boolean acceptForAnnotatorOrInspection(PsiElement element) {
...
+
+    public PsiElement getBeanPsiElement(PsiElement element) {
+        return enabledExtensions.stream()
+            .map(c -> c.getBeanPsiElement(element))
+            .filter(Objects::nonNull)
+            .findFirst().orElse(null);
+    }
+
+    public boolean isAnnotatedWithHandler(PsiMethod psiMethod) {
+        return  Arrays.stream(psiMethod.getAnnotations()).anyMatch(a -> a.getQualifiedName().equals("org.apache.camel.Handler"));
```
moelholm: Cannon use class ref ?

camel-idea-plugin/src/test/resources/testData/MyJavaBeanRenameResultTestData.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package testData;
+
+public final class MyJavaBeanTestData {
+
+    private MyJavaBeanTestData() { }
+
+    public void letNotDoThis() {
```
moelholm: lets

camel-idea-plugin/src/main/java/org/apache/camel/idea/completion/extension/CamelJavaBeanReferenceSmartCompletion.java
```diff
@@ -0,0 +1,132 @@
...
+            boolean isPrivate = getCamelIdeaUtils().isOneOfModifierType(method, JvmModifier.PRIVATE, JvmModifier.ABSTRACT);
+
+            if (!method.isConstructor() && !isPrivate) {
+                String presentableMethod = getPresentableMethodWithParameters(method);
+                answer.add(buildLookupElement(method, presentableMethod));
+            }
+        }
+        // are there any results then add them
+        if (!answer.isEmpty()) {
+            String hackVal = element.getText();
```
moelholm: Hack? :)

camel-idea-plugin/src/main/java/org/apache/camel/idea/completion/extension/CamelJavaBeanReferenceSmartCompletion.java
```diff
@@ -0,0 +1,132 @@
...
+        return  builder.withAutoCompletionPolicy(AutoCompletionPolicy.GIVE_CHANCE_TO_OVERWRITE);
+    }
+
+    private String getPresentableMethodWithParameters(PsiMethod method) {
+        final String parameters = getMethodParameters(method);
+        String presentableMethod = method.getName();
+        if (!parameters.isEmpty()) {
+            presentableMethod = String.format("%s(%s)", method.getName(), parameters);
+        }
+        return presentableMethod;
```
moelholm: Do you want to have an empty set of parantheses when no params are available?

camel-idea-plugin/src/main/java/org/apache/camel/idea/completion/extension/CamelJavaBeanReferenceSmartCompletion.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Return all method names for the specific class and it's super classes, except
+     * Object and Class
+     */
+    private Collection<PsiMethod> getMethods(PsiClass psiClass) {
+        return Stream.of(psiClass.getAllMethods())
+            .filter(p -> !p.isConstructor())
+            .filter(p -> !Object.class.getName().equals(p.getContainingClass().getQualifiedName()))
+            .filter(p -> !Class.class.getName().equals(p.getContainingClass().getQualifiedName()))
+            .collect(Collectors.toCollection(ArrayList::new));
```
moelholm: Consider toList()

bennetelli:
thanks for your thougts, @moelholm 👍

fharms:
Wow wow wow @moelholm did you just party crash my pull-request 🙂

moelholm:
He! Looks really good :)

camel-idea-plugin/src/main/java/org/apache/camel/idea/completion/extension/CamelJavaBeanReferenceSmartCompletion.java
fharms: if no parameter, the string "parameters" is empty :-)

camel-idea-plugin/src/main/java/org/apache/camel/idea/util/CamelIdeaUtils.java
fharms: No because we don't depend on camel core in the plugin code, and second it can't be a hard reference because it will throw a lot of class not found exception if the user has none Camel project open in IDEA

camel-idea-plugin/src/test/resources/testData/MyJavaBeanRenameResultTestData.java
fharms: Thanks :-)

davsclaus:
Maybe @moelholm is trying to tell @fharms that he wants to work on this plugin as well, and have all the same fun as you do with the jetbrains plugin api ;)

fharms:
@davsclaus I think you right and if I read between the lines it's a call from @moelholm for being part of something amazing

PR #388, opened 2018-02-21
corrected the hasQuestionMark check in stringutils

lhein:
Signed-off-by: Lars Heinemann <lhein.smx@gmail.com>


lhein:
0 is a valid return value of indexOf method and should be covered in that check method.

davsclaus:
Yeah you are essentially right, however its used for detecting `?` in camel endpoint uris so the `?` would never be in position 0, but lets fix it anyway. Thanks for spotting this and the PR.

Issue #386, opened 2018-02-18
Bean method completion in "when().method(<cursor>)


fharms:
Like we support bean method completion for "bean(MyClass.class,"<cursor>") we should also support it "when.method"

```
.choice()                
  .when().method(MyClass.class,"<cursor>")
.end()
           
```

fharms:
fixed in #386

PR #400, opened 2018-03-09
Bean method completion in "when().method(<cursor>) #386"

snurmine:
Handle method as bean . Not sure if this is the right way, but it seems to work in my Idea.


fharms:
@snurmine thanks for the contribution and it looks fine. 

Would you please add a few test cases to test the scenario with "method"?, you can see how it's done in the JavaCamelBeanReferenceSmartCompletionTestIT

snurmine:
Thanks, I added one test. Should this work in XML too ?

fharms:
Not at this moment, later we will add support for XML also

Again thanks for the contribution 👍

PR #401, opened 2018-03-10
Adding support for method completion when using bean field reference

fharms:


camel-idea-plugin/src/main/java/org/apache/camel/idea/service/extension/camel/JavaCamelIdeaUtils.java
```diff
@@ -170,10 +173,14 @@ public boolean acceptForAnnotatorOrInspection(PsiElement element) {
...
         if (beanPsiElement != null) {
-            PsiElement resolve = beanPsiElement.getReference().resolve();
-            if (resolve instanceof PsiClass) {
-                return (PsiClass) resolve;
+            PsiClass psiClass = null;
+            final PsiElement resolveElement = beanPsiElement.getReference().resolve();
+            if (resolveElement instanceof PsiClass) {
+                psiClass = (PsiClass) resolveElement;
+            } else if (resolveElement instanceof PsiField) {
+                psiClass = ((PsiClassReferenceType) PsiUtil.getTypeByPsiElement(resolveElement)).resolve();
```
davsclaus: Can there be a NPE if the getTypeByPsiElement returns null (if it can do so)

camel-idea-plugin/src/test/resources/testData/CompleteJavaBeanRoute6TestData.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.camel.main.Main;
+import testData.CompleteJavaBeanTestData;
+
+public final class CompleteJavaBeanRoute6TestData extends RouteBuilder {
+
+
+    @Override
+    public void configure() {
+        from("file:inbox")
+            .bean(beanTestData, "<caret>")
```
davsclaus: How can this code compile? Should it not be a .class type or something, as `beanTestData` is not a field

fharms: The purpose of the test is to test how the code behaves if there is an invalid field reference. So it's on purpose the beanTestData is not a field in this case

camel-idea-plugin/src/main/java/org/apache/camel/idea/service/extension/camel/JavaCamelIdeaUtils.java
fharms: good point

PR #404, opened 2018-03-11
Add real time jsonpath validation.

davsclaus:
Just a bit dirty prototype on this - needs a bit more polish. Also the unit tests with extra JARs on the classpath is a bit ugly with how to grab those mvn artifacts and turn them into idea libraries.

### DIFF
--- a/camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelJSonPathAnnotator.java
+++ b/camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelJSonPathAnnotator.java
@@ -23,7 +23,7 @@ import com.intellij.openapi.util.TextRange;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.xml.XmlAttributeValue;
 import org.apache.camel.catalog.CamelCatalog;
-import org.apache.camel.catalog.SimpleValidationResult;
+import org.apache.camel.catalog.LanguageValidationResult;
 import org.apache.camel.idea.service.CamelCatalogService;
 import org.apache.camel.idea.service.CamelPreferenceService;
 import org.apache.camel.idea.service.CamelService;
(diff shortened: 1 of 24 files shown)


davsclaus:
Gotta run, but you guys are welcome to take a look

camel-idea-plugin/src/main/java/org/apache/camel/idea/annotator/CamelJSonPathAnnotator.java
```diff
@@ -0,0 +1,134 @@
...
+                        String error = result.getShortError();
+                        if (error == null) {
+                            error = result.getError();
+                        }
+                        TextRange range = element.getTextRange();
+                        if (result.getIndex() > 0) {
+                            range = getAdjustedTextRange(element, range, text, result);
+
+                        }
+                        holder.createErrorAnnotation(range, error);
```
fharms: consider split this up in small methods, like "validateLanguage" ,"displayErrorAnnoation" eg.

camel-idea-plugin/src/main/java/org/apache/camel/idea/inspection/AbstractCamelInspection.java
```diff
@@ -152,6 +149,46 @@ private void validateSimple(@NotNull PsiElement element, final @NotNull Problems
...
+                }
+                if (!result.isSuccess()) {
+                    // favor the short error message
+                    String msg = result.getShortError();
+                    if (msg == null) {
+                        msg = result.getError();
+                    }
+                    holder.registerProblem(element, msg);
+                }
+            }
```
fharms: consider split this up in small methods, like "validateLanguage" ,"registerProblem" eg.

camel-idea-plugin/src/main/java/org/apache/camel/idea/service/extension/camel/JavaCamelIdeaUtils.java
```diff
@@ -92,7 +101,39 @@ public boolean isCameSimpleExpressionUsedAsPredicate(PsiElement element) {
...
+        PsiMethodCallExpression call = PsiTreeUtil.getParentOfType(element, PsiMethodCallExpression.class);
+        if (call != null) {
+            // okay dive into the psi and find out which EIP are using the simple
+            PsiElement child = call.getFirstChild();
+            if (child instanceof PsiReferenceExpression) {
+                PsiExpression exp = ((PsiReferenceExpression) child).getQualifierExpression();
+                if (exp == null) {
+                    // okay it was not a direct method call, so see if it was passed in as a parameter instead (expression list)
+                    element = element.getParent();
+                    if (element instanceof PsiExpressionList) {
```
fharms: Since it's not an else if condition, I guess the goal is to get the PsiMethodCallExpression.class ?

If that's the case it should be fine you just use `exp  = PsiTreeUtil.getParentOfType(element,PsiMethodCallExpression.class)`, since it will travers up in the tree until it find a parent instance of PsiMethodCallExpression.class.

fharms:
Nice work 👍

camel-idea-plugin/src/main/java/org/apache/camel/idea/service/extension/camel/JavaCamelIdeaUtils.java
davsclaus: Maybe but its a copy of the existing working code we have for the simple language, and this should be identical, so I wont risk changing it.

fharms: If it's a copy of existing code we should move it out to a common class for sharing

fharms: If it's a copy of existing code we should move it out to a common class for sharing

PR #423, opened 2018-03-29
Fix for find usage was not working from bean method to Camel bean DSL

fharms:
This fix an issue with the CamelBeanMethodReference was used
incorrectly, and referer to the bean method and not the PsiLiteral bean DSL.
The result was when using "find usage" it just pointed to the
method itself and not the Camel Bean DSL


davsclaus:
Adding a little note to the changelog in `plugin.xml` would be good for this new great stuff

PR #424, opened 2018-04-03
Camel bean annotator

fharms:
- Annotate camel bean method reference with an error if the method
is private or not exist.
- Spring clean up and refactoring around the code shared in the
smart completion, element reference code and the annotator


davsclaus:
In the settings UI it would be nice if there is an option to turn this on | off. We have that for the other annotators.

davsclaus:
Add a note to the changelog in the plugin.xml

PR #458, opened 2018-07-22
fix NPE in JavaIdeaUtils

bennetelli:


bennetelli:
Not sure if this fixes the initial error, but it will fix the NPE at least

PR #476, opened 2019-03-19
Depdency upgrades in maven plugins

davsclaus:


davsclaus:
Okay CS update has a bunch of new rules which needs to be fixed first

PR #489, opened 2019-05-05
Springbean support

fharms:
This extend the bean code completion and find usage with support also for Spring beans


camel-idea-plugin/src/test/resources/testData/completion/method/CompleteJavaSpringComponentBeanTestData.java
```diff
@@ -0,0 +1,29 @@
+package testData;
+import org.springframework.stereotype.Component;
+
+/**
```
davsclaus: The licenseheder is in the top of the file, before package and imports

PR #490, opened 2019-05-05
Migrate to gradle build

fharms:


fharms:
@davsclaus yeah it seem a lot easier with the gradle plugin and it also support development in IU and publish plugin from gradle

fharms:
@davsclaus I renamed the package in this commit 5983a965e0d5f5f88a3aa3493ed13677237a945c

PR #515, opened 2019-08-12
Upgrade 2019 2

fharms:


camel-idea-plugin/src/test/java/com/github/cameltooling/idea/gutter/BeanInjectLineMarkerProviderTest.java
```diff
@@ -74,7 +74,7 @@ private void validateBeanInjectWithMultipleTargets(RelatedItemLineMarkerInfo<? e
 
     private void validateBeanInjectWithValue(RelatedItemLineMarkerInfo<? extends PsiElement> marker) {
         List<String> targetBeanNames = getTargetBeanNames(marker);
-        assertEquals(1, targetBeanNames.size());
+        assertEquals(2, targetBeanNames.size());
```
rastislavpapp: I'm not sure about this, this would mean that the reference to `testClass2Bean` in ```@BeanInject("testClass2Bean")``` would also be a reference to `testClass2Bean2`, which is completely different bean, with a simillar name :-/

rastislavpapp: I see that IDEA API changed in `BeanInjectLineMarkerProvider#getBeanReference`. This code should be equivalent to the previous implementation:

```
return Optional.ofNullable(beanInjectAnnotation.findAttributeValue("value"))
                .map(v -> Arrays.stream(v.getReferences()))
                .orElse(Stream.empty())
                .filter(r -> r instanceof BeanReference)
                .map(r -> (BeanReference) r)
                .findFirst()
                .orElse(null);
```

rastislavpapp: I've created a pull request from my fork to your fork for this (https://github.com/fharms/camel-idea-plugin/pull/1) - I don't know if this is the proper procedure for this situation :-)

fharms: Pull request merged. Thanks @rastislavpapp

PR #637, opened 2021-12-19
Debugger - Resolves #98

### LINKED ISSUE #98: Camel route debugger
This is maybe a bit more hardcore to implement. But would be nice if we can add breakpoints in the route dsl, and then have IDEA step through the route in debug mode.

javaduke:
This is the first iteration of the Camel route debugger implementation. Current limitations are:

- Can debug XML routes only, no Java or other DSLs;
- requires Camel 3.15 or later;

Please review and let me know if there are any fixes or improvements required.
Linked issue: #98


davsclaus:
There is a lot of code, so I can see that you have runners to run camel in debugging mode - since Camel can run anywhere, then I am not too keen on having specific runners that are tied to spring boot only and its only for debugging.

Instead users should set that option to enable camel debugging. We can look at adding a camel-debug JAR as dependency that users can add to their project, that Camel then detects on startup and runs in debug mode.

I assume you will continue to polish and cleanup the code as there is a bunch of commented out code, from Mule ;)

The language expression for simple/datasonnet - I wonder if IDEA think that its a programming language with its own file extensions ala groovy, kotlin etc. I wonder if that is needed or a hack, and if maybe there is another way to hook into the idea debugger to do remote evaluations - but that can be researched later.

It would have been good if this PR was not a giant PR, but was broken up into different tasks. I suggested you to focus on the debugger, and then add expression evaluation later.

Its a massive code to take in to support, but an awesome feature, so keep on hacking. And lets get it not tied to spring boot only, and later see if we can do other languages than XML. And add more expression evaluations for more Camel languages like jsonpath.

davsclaus:
Also I wonder if we can make this work for non maven projects - or at least isolate what parts that are tied to maven, so we can look at supporting gradle users too. But that can be for another ticket.

davsclaus:
Yes it should be named `camel.sprinboot.debugging` when using spring-boot, and `camel.main.debugging` for other runtimes with camel-main (standalone, camel-k, or quarkus)

davsclaus:
Screenshot 2021-12-20 at 15 44 40

davsclaus:
Great work - looks fantastic with the debugger, and the stepping is fast.

I agree enabling debugging mode is a bit cumbersome

- camel-management JAR must be on classpath
- camel.*.debugging = true

I think we could do a new camel-debug JAR which you just add as dependency, then it pulls in camel-management and enables debugging mode, and also do a INFO logging that debugging is enabled so you can see it when starting up.

davsclaus:
It would be good to have _Set Value_ supported in the debug panel for the message body, and then later for the headers / properties.

javaduke:
It is my understanding that custom runners are required, otherwise the debugger will not know if it is supposed to run or not (that is, we don't want to enable it for non-Camel projects). But I will see if I can find a better way and remove the custom runners.

For language expressions I had to define IDEA Language objects, but I can add other languages too, unfortunately, each one has to be created explicitly. But there's no need for custom extensions or file types, the `Language` classes are enough. Perhaps I will add a static class `CamelLanguages` listing them all and we can add new languages as needed.

I'll work on setValue functionality, but there are still a few quicks and features to implement, so I will add it to my backlog. I'll create Jiras and assign them to myself.

I'll clean up the code to remove any Mule references. As you may have guessed, I reused some portions of the Mule plugin I wrote a few years ago :)

davsclaus:
Sounds good, I have `camel-debug` ready tomorrow - will add a `camel-debug-starter` also for spring boot so its easy to enable by just adding this JAR.

Okay about the runners, we can look at this later. You can create the TODO tasks as issues in this project - I will invite you in the org so you can work on it more easily. Thanks for the help.

javaduke:
Thank you, Claus! Really excited to work on this with you!
So what's the difference between `camel-management` and `camel-debugger`?
The problem with these is that it is still necessary to add it manually to the POM, and it is not required for production, so needs to be removed before, say, merging to `master`. Also, I noticed that in order to retrieve message dumps, one must also have `camel-xml-jaxb` in the classpath. And if I want to evaluate expressions in DataSonnet, or any other language that is not part of the core, I must have those in the classpath too.
I'm trying to find a way to add these dependencies "on the fly", just for the duration of the debugging process, but unfortunately I don't get much help from the IntelliJ team :(

davsclaus:
Okay I can get `camel-debug-starter` to automatic turn on debugging and start the debugger with `enableDebug` so its ready, however I guess your PR needs still to do some custom "thingy" to work - so we may need more work.

The advantage with `camel-debug` is that we can add other needed dependencies such as `camel-xml-jaxb` to have XML dump of messages.

You can have `camel-debug` as test scope in the project, then its at least downloaded and resolved by maven.

davsclaus:
I am using this example that has `camel-debug` included
https://github.com/apache/camel-spring-boot-examples/tree/main/xml

davsclaus:
You can potentially try with setting scope=provided for `camel-debug`

javaduke:
OK, I think the first version can be released. I'm going to work on additional features later.

PR #647, opened 2021-12-24
Implements SetValue - resolves #642

### LINKED ISSUE #642: Debugger - implement setValue
There should be an additional action to set value of the message body and/or properties and headers.

javaduke:


camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/actions/CamelSetValueAction.java
```diff
@@ -0,0 +1,61 @@
...
+        return setValueActionHandler;
+    }
+
+    @Override
+    public void update(@NotNull final AnActionEvent event) {
+        super.update(event);
+        if (event.getPresentation().isEnabledAndVisible()) {
+            //If we are debugging Camel project, this should be enabled - but how do we know it's a Camel project?
+            //Find Maven project and see if there's a Camel dependency
+            MavenProject mavenProject = MavenActionUtil.getMavenProject(event.getDataContext());
```
davsclaus, line 49: You can use CamelService which can tell this
com.github.cameltooling.idea.service.CamelService#isCamelPresent

javaduke: Good idea, thanks, done!

davsclaus (approved):

Issue #650, opened 2022-01-02
Debugger - support other DSLs


javaduke:
The first iteration should have support for Java and YAML DSLs.

PR #651, opened 2022-01-03
Java DSL support

### LINKED ISSUE #645: Debugger - Able to run via camel-debug on classpath
We have special runners which runs the app via maven and its own maven goals.

We should be able to debug via IDEA normal debugger runner, by having camel-debug on classpatn then camel-core is automatic in debug mode

This example has been prepared for this, and can be used as model for this ticket
https://github.com/apache/camel-spring-boot-examples/tree/main/xml

### LINKED ISSUE #649: Debugger - support breakpoints inside interceptors

javaduke:
- Improved Java DSL support per Issue #650;
- Dedicated run configurations are no longer required (however, SpringBoot apps must run with `-Dspring-boot.run.fork=false`;
- Resolves #649;
- Resolves #645;

### DIFF
--- a/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
+++ b/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
@@ -33,13 +33,10 @@ import com.intellij.psi.JavaPsiFacade;
 import com.intellij.psi.PsiClass;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiFile;
-import com.intellij.psi.PsiManager;
 import com.intellij.psi.search.FilenameIndex;
 import com.intellij.psi.search.GlobalSearchScope;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.psi.xml.XmlTag;
-import com.intellij.psi.xml.XmlToken;
-import com.intellij.psi.xml.XmlTokenType;
 import com.intellij.xdebugger.AbstractDebuggerSession;
 import com.intellij.xdebugger.XDebugSession;
 import com.intellij.xdebugger.XDebuggerUtil;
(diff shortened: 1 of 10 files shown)


davsclaus (approved):

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/JavaCamelIdeaUtils.java
```diff
@@ -53,6 +56,35 @@
...
+                        "RouteBuilder",
+                        "RoutesBuilder",
+                        "RouteConfigurationBuilder",
+                        "RouteConfigurationsBuilder",
+                        "AdviceWithRouteBuilder",
+                        "EndpointRouteBuilder",
+                        "EndpointRouteConfigurationBuilder",
+                        "SpringRouteBuilder"
+                });
+        if (file != null && file.getFileType().equals(JavaFileType.INSTANCE)) {//This is Java file
```
apupier, line 72: better practice to invert equals to avoid potential NPE in case getFileType returns null

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
```diff
@@ -103,11 +98,14 @@
     private ObjectName debuggerMBeanObjectName;
 
     private org.w3c.dom.Document routesDOMDocument;
+    //private org.w3c.dom.Document locationsDOMDocument;
```
apupier, line 101: commented out code can be removed

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
```diff
@@ -308,8 +317,14 @@ public void stepOver(XSourcePosition position) {
...
+
+            if (breakpointElement == null) {
+                //TODO Log warning
+                return;
+            }
+            String breakpointId = breakpointElement.keySet().iterator().next();
+
             //Get the route of the current tag
             Element routeElement = getParentRouteId(breakpointId);
             String routeId = routeElement.getAttribute("id");
```
apupier, line 330: it is checked for nullity but not for emptiness?

If I understand well the createBreakpointElementFromPosition(position); is expected to create a Map of a single element each time? Won't it be better to provide a specific data structure for it?

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
```diff
@@ -393,6 +411,7 @@ private void nextStep(XSourcePosition position, boolean isOver) {
                 resume();
             }
         } else {
+//            backlogDebugger.step();
```
apupier, line 414: commented out code can be removed.

it reveals an interesting poitn which is that the backlogDebugger;step is not used to a step action. This is somethign used in Eclipse https://github.com/jbosstools/jbosstools-fuse/blob/dd8b033accdcbb8f35d0f6c8aab5ce18e35730fc/editor/plugins/org.fusesource.ide.launcher/src/org/fusesource/ide/launcher/debug/model/CamelThread.java#L308

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
javaduke: Yes, I thought about it, but it is not used anywhere else, so I decided just to use a typed singleton map. And the way it's created it cannot be empty, either a null or containing a single key-value pair.

PR #658, opened 2022-01-11
Minor improvements

javaduke:
- DataSonnet Language Injection;
- camel-debug detection in classpath.


davsclaus:
There are checkstyle failures when you change the order of the imports.
These errors you only get when running the build with gradle.

Wonder if you could fix these CS errors

javaduke:
Strange, I just ran checkstyle and see no errors...

davsclaus:
Did you run from command line

```
./gradlew clean build
```

There is reports the checkstyle errors

PS: Yes there are test errors

javaduke:
Yes, just did it again, no errors. I also ran `gradlew checkstyleMain`, no errors.

PR #683, opened 2022-05-05
Ref #644: Yaml DSL support

### LINKED ISSUE #644: Yaml DSL support
Add support for code completion et all when using yaml dsl

essobedo:
Fix #644 

## Motivation

Yaml DSL is not yet supported in the plugin

## Modifications:

* Implements all the required methods in `YamlCamelIdeaUtils` to support gutter, endpoint validation, simple and json path validation
* Adds support of completion of header name and value
* Renames `CamelPropertiesOrYamlFileReferenceContributor` to `CamelPropertiesFileReferenceContributor` since we need a dedicated class for Yaml
* Adds `CamelYamlFileReferenceContributor` to define all the `CompletionProvider` for Yaml files
* Modifies `CamelHeaderValueCompletion` to take into account all the specificities of Yaml elements
* Updates the read me
* Migrates the IT `CamelInspectXmlEndpointTestIT` to use it as example for Yaml (not related to the initial issue)

### DIFF
--- a/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/YamlCamelIdeaUtils.java
+++ b/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/YamlCamelIdeaUtils.java
@@ -16,27 +16,37 @@
  */
 package com.github.cameltooling.idea.service.extension.camel;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Predicate;
+
 import com.github.cameltooling.idea.extension.CamelIdeaUtilsExtension;
+import com.github.cameltooling.idea.util.IdeaUtils;
+import com.github.cameltooling.idea.util.YamlPatternConditions;
 import com.intellij.ide.plugins.IdeaPluginDescriptor;
 import com.intellij.ide.plugins.PluginManagerCore;
+import com.intellij.openapi.components.ServiceManager;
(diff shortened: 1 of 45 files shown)


camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/YamlCamelIdeaUtils.java
```diff
@@ -47,6 +57,130 @@ public class YamlCamelIdeaUtils extends CamelIdeaUtils implements CamelIdeaUtils
             "route",
             "routeConfiguration"
         });
+    /**
+     * All keys representing the potential producers.
+     */
+    private static final String[] PRODUCERS = {
+        "to", "tod", "intercept-send-to-endpoint", "interceptSendToEndpoint", "wire-tap", "wireTap",
```
davsclaus, line 64: to dynamic is also `toD` in yaml dsl

davsclaus (approved):
Awesome work

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/YamlCamelIdeaUtils.java
essobedo: good catch, there is also `to-d`

essobedo: Fixed

PR #686, opened 2022-05-10
Ref #684: Option name and value completion in properties files

### LINKED ISSUE #684: Option name and value completion in properties files
It would be good to have support for end users configuring their application.properties

All the main config is possible to use,
https://camel.apache.org/components/3.16.x/others/main.html

```
camel.main.messageHistory=true
```

Also this works for all the components, so you can do

```
camel.component.seda.queueSize=50
```

We have all the options listed in camel-catalog so it should be possible to find out which options are available, what type they are, default values, enums and so on. Like we do for endpoint uri completions.

essobedo:
fixes #684

## Motivation

The completion of option name and value in properties file is not yet supported.

## Modifications

* Add `CamelPropertyKeyCompletion` to manage the completion of property key covering options of main, data format, component and language
* Add `CamelPropertyValueCompletion` to manage the completion of property value covering options of main, data format, component and language
* Avoid getting the catalog in a loop (not related to the current issue)
* Fix typos in existing code (not related to the current issue)


camel-idea-plugin/src/main/java/com/github/cameltooling/idea/completion/contributor/CamelPropertiesFileReferenceContributor.java
```diff
@@ -29,10 +34,26 @@ public class CamelPropertiesFileReferenceContributor extends CamelContributor {
...
                 new EndpointCompletion(getCamelCompletionExtensions())
         );
+        // The key of the property corresponding to key characters in a properties file
+        extend(CompletionType.BASIC,
+            psiElement(PropertiesTokenTypes.KEY_CHARACTERS)
+                .inside(PsiFile.class)
+                .inFile(filePattern),
+            new CamelPropertyKeyCompletion()
+        );
+        // The key of the property corresponding to value characters in a properties file
```
davsclaus, line 50: Should it be - The value of ...

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/completion/property/CamelPropertyValueCompletion.java
```diff
@@ -0,0 +1,355 @@
...
+        if (propertyKey == null) {
+            return;
+        }
+        final Optional<? extends BaseOptionModel> option = getOption(element, propertyKey);
+        if (option.isEmpty()) {
+            return;
+        }
+        final List<LookupElement> answer = getSuggestions(option.get());
+        if (!answer.isEmpty()) {
+            // sort the headers A..Z which is easier to users to understand
```
davsclaus, line 80: sort the headers ... sort the values

davsclaus (approved):
Yet another great work

davsclaus:
Does it work with showing the docs for each option if you press `ctrl + j` I think is. the key combo, that is neat when you scroll up/down the popup where a 2nd window shows the doc side-by-side. 

Something that uri endpoint can do.

essobedo:
> Does it work with showing the docs for each option if you press `ctrl + j` I think is. the key combo, that is neat when you scroll up/down the popup where a 2nd window shows the doc side-by-side.
> 
> Something that uri endpoint can do.

No it doesn't on properties file

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/completion/contributor/CamelPropertiesFileReferenceContributor.java
essobedo: good catch thx

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/completion/property/CamelPropertyValueCompletion.java
essobedo: thx

PR #696, opened 2022-05-19
Ref #689: Allow to detect the Camel Runtime

### LINKED ISSUE #689: Allow to detect the Camel Runtime
There are several existing Camel Runtimes and for each of them the Camel options' key and values set in properties files can change so it could be interesting to be able to detect the Camel Runtime and adapt the suggestions accordingly.

For example, in case of a Camel Spring Boot project, we would expect the property key to start with `camel.springboot.` instead of `camel.main.`.

essobedo:
fixes #689 

## Motivation

There are several existing Camel Runtimes and for each of them the Camel options' key and values set in properties files can change so it could be interesting to be able to detect the Camel Runtime and adapt the suggestions accordingly.

## Modifications:

* Add the ability to set the Camel Runtime in the configuration
* Force a reset of the catalog when the Camel Runtime configured has changed
* Adds an enumeration called `CamelCatalogProvider` to list all the supported Camel Runtime with their specificities
* Adds a specific `JSonSchemaResolver ` called `SpringBootJSonSchemaResolver` to be able to leverage the Spring configuration metadata to adapt the main and component schemas
* Adapts the code of  `CamelPropertyKeyCompletion` and `CamelPropertyValueCompletion` to support keys that don't start with `camel` to be able to support keys of type `management.endpoint.camelroutes.cache.time-to-live`
* Enables Camel options' key and value completion only when Camel is detected

## Result:

**Configuration of the Camel Runtime:**

35-configuration-camel-runtime

**Completion adapted to the Camel Spring Boot Runtime:**

36-completion-spring-boot-option-name-properties


davsclaus (approved):

davsclaus:
Can you add to `plugin.xml` a few lines for its changelog, as this is what is shown by IDEA when users upgrade

davsclaus:
Awesome, love that we can now do this for different runtimes, great work on improving this Camel tool

essobedo:
> Can you add to `plugin.xml` a few lines for its changelog, as this is what is shown by IDEA when users upgrade

Done

PR #699, opened 2022-05-25
Ref #690: Add code completion in yaml files

### LINKED ISSUE #690: Add code completion in yaml files
Spring Boot and Quarkus users are use to configure their application with either a file `application.properties` or `application.yml`. As we now have completion supported in file `application.properties`, the goal of this task is to support the completion also in YAML files.

essobedo:
fixes #690

## Motivation

Quarkus and Spring Boot users can configure their application using yaml files, so it could be helpful to propose code completion on yaml files in case of those specific Camel runtimes.

## Modifications:

* Propose 2 different implementation of `CamelPropertyKeyCompletion` one for properties file and the other one for yaml files
* Adapt the code of `CamelPropertyKeyCompletion` to be able to propose only subpart of the original suggestion or to allow to ignore some suggestions, to be compatible with a yaml file
* Add a static inner class `SuggestionContext` to bring more context when creating the `LookupElementBuilder` instances
* Enable the completion for Yaml files only when camel is present, the camel runtime is Quarkus or Spring Boot and the file is not the definition of a route
* Add a new utility method in `YamlPatternConditions` called `withPrevSibling` to define a matching pattern on previous sibling
* Add a new utility method in `YamlIdeaUtils` called `getIndent` to compute the indent 
* Improve the way to auto indent the header value in case of a yaml file by using the method `CodeStyleManager #adjustLineIndent` (not related to the issue)

## Result

**Property key completion**

37-completion-option-name-yaml

**Property value completion**

38-completion-option-value-yaml


davsclaus (approved):
Yet again awesome - Makes Camel more modern with yaml configuration support - love that I can do this now - thanks

Issue #714, opened 2022-06-14
Remove kameletName in kamelet drop down box


davsclaus:
In the left hand side you can see "KameltName" in the drop down box for each listed kamelet
https://github.com/camel-tooling/camel-idea-plugin/blob/main/img/39-completion-kamelet-name.png

Can we remove this so its empty.

PR #716, opened 2022-06-15
Ref #713 - Add Kamelets' icon in suggestions and description in the quick doc

### LINKED ISSUE #713: kamelets - Can we show small icon per kamelet in the drop down box
In this image we list the known kamelets
https://github.com/camel-tooling/camel-idea-plugin/blob/main/img/39-completion-kamelet-name.png

We have a Camel icon there. But wonder if we could maybe use the icons for each kamelet as they have their own icons.
Well they are embeded in the kamelet yaml file as SVG blurb. So not sure if an in-memory image or something can be built for IDEA to use to show.

essobedo:
fixes #713 and #714

## Motivation

Each Kamelet has its own icon and should be proposed when suggesting the name of Kamelets.

## Modifications:

* Improve the `KameletService` to retrieve the content of the icon from the Kamelet definition and to load it with a specific size
* Adapt the code to provide a specific icon to a context path suggestion
* Use a pool of Icons to reuse the same instance of `Icon` when possible to limit the impact on the heap

## Result

39-completion-kamelet-name


davsclaus (approved):
Yay awesome

PR #725, opened 2022-06-27
Fix warnings in the debugger package

essobedo:
## Motivation

While investigating in how the recent improvements in the Debugger of Apache Camel can be leveraged in the IntelliJ plugin, I realized that many classes in the package `debugger` has warnings raised by IntelliJ. Moreover there are many line of codes that are commented which make the code very hard to read. To improve the maintainability of this package, it could be nice to address the warnings raised by IntelliJ.

## Modifications:

* Use the logger instead of calling `e.printStackTrace()`
* Use `final` keyword on fields that are not modified which prevents visibility issues
* Use lambda when possible to improve the readability of the code
* Avoid setting the size to the array when copying a collection into an array
* Remove commented code
* Replace deprecated code when possible
* Inline the code instead of creating a variable for a single usage
* Fix the generic related issues
* Add a log message in catch blocks for debugging purpose instead of keeping them empty which could hide bugs
* Always call `Thread.currentThread().interrupt()` when catching `InterruptedException`
* Improve the code calling `dumpTracedMessagesAsXml` by calling the old signature only as fallback instead of always calling it potentially for nothing
* Catch error also for pending breakpoints to remove
* Call `Thread.onSpinWait()` to improve the implementation of a busy wait
* Call static methods directly from the class `IdeaUtils` instead of through an instance.
* Remove unused variables, methods, fields and parameters
* Simplify the initialization of `NO_BREAKPOINTS_AT` by relying on the varargs parameter of the method `Arrays.asList`
* Avoid reading several times the `volatile` field `myDelegate` for the same call to ensure consistency.
* Implement the double-checked locking idiom to initialize the `volatile` field `myDelegate`
* Avoid adding a field when a local variable is good enough

### DIFF
--- a/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebugProcess.java
+++ b/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebugProcess.java
@@ -27,6 +27,7 @@ import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.DefaultActionGroup;
 import com.intellij.openapi.actionSystem.Presentation;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.util.ArrayUtil;
 import com.intellij.xdebugger.XDebugProcess;
 import com.intellij.xdebugger.XDebugSession;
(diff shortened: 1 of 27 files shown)


essobedo:
@javaduke Hi, if you have the opportunity to review it, it would be very helpful as you seem to be the most qualified in this part of the code so you are the best person to review it and to share what you think of it. Thank you in advance for any help 🙏

javaduke:
Wow, looks really good, thank you for working on it! Unfortunately it would take me a long while to review it all, but I understand the changes are primarily cosmetic, so as long as nothing is broken, let's merge it :)

essobedo:
@javaduke yeah I confirm that the logic of the code has not been changed but do you know what I can do to ensure that nothing is broken? Are the tests only manual? and if so, could you please tell me what should I test? So far, I only tested the nominal use case (Camel break points + Camel Application Debug Config)

javaduke:
Yes, I'm afraid the tests are only manual at this point. Just test everything, step into, step over, step out, evaluate expressions in Simple and DataSonnet, Java and XML routes, etc.

essobedo:
@davsclaus Any suggestions, remarks, comments about this PR before merging it?

davsclaus (approved):

PR #728, opened 2022-07-05
Ref #727 - Make the Camel Debugger UX easier

### LINKED ISSUE #727: Make the Camel Debugger UX easier
it would be great if we can make the camel-debug UX easier and better in IDEA so its easier to click "debug" or whatever you need and that it works on camel-spring-boot / camel-main / and potentially camel-quarkus in a similar/easy way.

essobedo:
fixes #727

## Motivation

it would be great if we can make the camel-debug UX easier and better in IDEA so its easier to click "debug" or whatever you need and that it works on camel-spring-boot / camel-main / and potentially camel-quarkus in a similar/easy way.

## Modifications:

* Adds a new enum `CamelRuntime` for all the supported Camel runtimes as it is now needed for the Camel Debugger and for the `CamelCatalogProvider` and rewrites the code of the `CamelCatalogProvider` consequently
* Adds a new check box in the preferences to be able to enable or not the auto set up of the Camel Debugger
* Re-writes the "Camel Runner" related code to avoid code duplication and to easily add a new one for Quarkus
* Removes the "before run tasks" related code as it is not used (not directly related to the original issue) 
* Simplifies the code in the debugger package to remove unused code 
* Fixes thread safe issues (not directly related to the original issue) 
* Improves the JMX related code to invoke the new attach and detach methods, to properly close the connection and to be able to connect from an URI for Quarkus as it is launched in another process.
* Replaces the maven goal `package` with `compile` to avoid having to deal with the tests
* Removes the `debugging` and `messagehistory` options as they are already set programmatically by Camel when the Camel Debugger is detected
* Adds the ability to reconnect to support Quarkus hot reload 
* Rewrites the method allowing to connect with retries as it filled up the stack (avoid recursion to prevent StackOverflow)

## Result

The artifacts camel-debug and camel-management are automatically added to a temporary pom file in case of the application is launched via a Camel runner. The temporary pom file is automatically removed when stopping the debug process. 

The artifacts camel-debug and camel-management are automatically downloaded and added to the classpath in case of the application is launched via Java runner unless it is a Quarkus application as for now it is hardly doable.

Examples of result:

44-camel-debugger-springboot-camel-launcher

45-camel-debugger-springboot-java-launcher

46-camel-debugger-springboot-java-test-launcher


apupier (commented):
I'm really not familiar with IntelliJ codebase so pick my comment for what it worth.

Based on the animated image, it seems to really simplify the usage of it.


Camel 3.18 has been released since you created the PR (given the delay before I provided feedback on it sorry). I think it is worthy to update the PR to not rely on the snapshot version.

Using the JMX url instead of the local process to connect opens possibilities for remote connection in the future.

It is leveraging the attach/detach which is making the debugging a lot more reliable for routes that are starting automatically.

essobedo:
@apupier Thx for the comment, I have upgraded the version of Camel within the context of the Hot reload support of Quarkus

davsclaus (approved):

davsclaus:
Awesome work

PR #733, opened 2022-07-08
Ref #732 - Improve the multi-breakpoints support of the Camel Debugger

### LINKED ISSUE #732: Instability when several breakpoints are set
If we set several breakpoints or we set one breakpoint and we invoke `step over` command, in case we have a route with regular incoming messages, the debugger keeps on switching from different active breakpoints which makes the debugging hardly understandable.

See below what we can face without doing anything in such use case:

problem-with-multi-breakpoints

essobedo:
fixes #732 

## Motivation

If we set several breakpoints or we set one breakpoint and we invoke `step over` command, in case we have a route with regular incoming messages, the debugger keeps on switching from different active breakpoints which makes the debugging hardly understandable.

## Modifications:

* Processes the suspended breakpoint node ids as batch
* Keeps only one Camel message per batch and ideally the one corresponding to the latest message processed
* Prevents getting `IndexNotReadyException` when collecting message info
* Sorts the messages by timestamp to get a more predictable behavior
* Adds the exchange Id as variables accessible from the debugger view

## Result

The debugger doesn't switch anymore from one breakpoint to another.


essobedo:
Depends on https://github.com/camel-tooling/camel-idea-plugin/pull/731

davsclaus (approved):

PR #771, opened 2022-11-18
Ref #769: Allow to load a runtime provider w/o the core artifacts

### LINKED ISSUE #769: Allow to load a runtime provider without the core artifacts
When we force a runtime by settings if the corresponding core artifacts are not part of the dependencies of the project, the catalog is not downloaded while we would expect to download it anyway.

essobedo:
fixes #769 

## Motivation

When we force a runtime by settings if the corresponding core artifacts are not part of the dependencies of the project, the catalog is not downloaded while we would expect to download it anyway.

## Modifications:

* Add a fallback method that will retrieve the catalog corresponding to the version of Camel
* Do nothing in the case of Quarkus as the versions are different
* Add a specific `grapeConfig.xml` as a temporary workaround to prevent dependency retrieval issue

### DIFF
--- a/camel-idea-plugin/grapeConfig.xml
+++ b/camel-idea-plugin/grapeConfig.xml
@@ -1,0 +1,35 @@
+<!--
+
+     Licensed to the Apache Software Foundation (ASF) under one
+     or more contributor license agreements.  See the NOTICE file
+     distributed with this work for additional information
+     regarding copyright ownership.  The ASF licenses this file
+     to you under the Apache License, Version 2.0 (the
+     "License"); you may not use this file except in compliance
+     with the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+     Unless required by applicable law or agreed to in writing,
+     software distributed under the License is distributed on an
+     "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+     KIND, either express or implied.  See the License for the
+     specific language governing permissions and limitations
+     under the License.
+
+-->
+<ivysettings>
+  <settings defaultResolver="downloadGrapes"/>
+  <resolvers>
+    <chain name="downloadGrapes" returnFirst="true">
+      <filesystem name="cachedGrapes">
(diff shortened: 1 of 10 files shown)


camel-idea-plugin/grapeConfig.xml
```diff
@@ -0,0 +1,36 @@
...
+    <chain name="downloadGrapes" returnFirst="true">
+      <filesystem name="cachedGrapes">
+        <ivy pattern="${user.home}/.groovy/grapes/[organisation]/[module]/ivy-[revision].xml"/>
+        <artifact pattern="${user.home}/.groovy/grapes/[organisation]/[module]/[type]s/[artifact]-[revision](-[classifier]).[ext]"/>
+      </filesystem>
+      <!-- Remove localm2 as temporary fix to prevent getting dependency retrieval issues with Grape -->
+      <!--ibiblio name="localm2" root="${user.home.url}/.m2/repository/" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true"/-->
+      <!-- TODO: add 'endorsed groovy extensions' resolver here -->
+      <ibiblio name="ibiblio" m2compatible="true"/>
+      <ibiblio name="jcenter" root="https://jcenter.bintray.com/" m2compatible="true"/>
```
davsclaus, line 33: I think jcenter is closed, so I think it should be removed
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

davsclaus (approved):

camel-idea-plugin/grapeConfig.xml
essobedo: Thx for the remark, TBH I was aware of it but by default, I initially decided to apply the changes on the default configuration for this particular version of Groovy which is https://github.com/apache/groovy/blob/486d57f832a577ad2c574868e3540ebe21c44abb/src/resources/groovy/grape/defaultGrapeConfig.xml
But you are right, maybe it is better to remove it even if it is only used for the tests

essobedo:
@dependabot squash and merge

Issue #788, opened 2022-12-28
IDEA 2022.3.x support


davsclaus:
This version has breaking changes so its a bit tougher to support

@essobedo I pushed the build files to upgrade to latest IDEA 2022.3.x.
The older 2022.1 and 2022.2 works, but 2022.3 has many compile error. So it seem JetBrains did some bigger changes there.
I still think we need to move forward, so we may end up with 2022.3 as the minimum version if we cannot make it work with older versions as well.

essobedo:
Fixed by https://github.com/camel-tooling/camel-idea-plugin/pull/801

PR #801, opened 2023-03-08
Support for IntelliJ 2022.3.2

ammachado:


build-all.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-for v in "2022.1.4" "2022.2.4" "2022.3.1"; do
+for v in "2022.1.4" "2022.2.4" "2022.3.1" "2022.3.2"; do
```
essobedo, line 3: Only the last bug fix version of each supported minor versions should be tested, so could you please update 2022.3.1

essobedo:
I forgot that with this version, it doesn't compile anymore, it is a known issue https://github.com/camel-tooling/camel-idea-plugin/issues/788

ammachado:
Updated plugin version and JDK, as per this [blog post](https://blog.jetbrains.com/platform/2022/08/intellij-project-migrates-to-java-17/).

essobedo:
I'm afraid that migrating to Java 17 is more complicated than this, there are breaking changes so we will need to drop support of 222 versions.
Have you tried to launch the tests locally with your changes?

essobedo (approved):
It looks like it works properly, thx for the PR 👍

PR #802, opened 2023-03-08
Refactored the majority of warnings reported on `runPluginVerifier`

ammachado:


camel-idea-plugin/src/main/java/com/github/cameltooling/idea/reference/CamelBeanReferenceContributor.java
```diff
@@ -119,15 +118,11 @@ private PsiReference[] createSpringBeanMethodReference(@NotNull PsiElement eleme
...
-    }
-
-    private JavaMethodUtils getJavaMethodUtils() {
-        return ServiceManager.getService(JavaMethodUtils.class);
+        return CamelIdeaUtils.getService();
     }
 
     private JavaClassUtils getJavaClassUtils() {
-        return ServiceManager.getService(JavaClassUtils.class);
+        return ApplicationManager.getApplication().getService(JavaClassUtils.class);
```
essobedo, line 125: ditto

camel-idea-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -47,6 +47,7 @@
 
   <extensions defaultExtensionNs="com.intellij">
     <notificationGroup id="Apache Camel" displayType="BALLOON"/>
+    <postStartupActivity implementation="com.github.cameltooling.idea.CamelPluginStartup"/>
```
essobedo, line 50: What is the benefit of it versus what we currently do?

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/annotator/AbstractCamelAnnotator.java
```diff
@@ -40,6 +40,9 @@
  */
 abstract class AbstractCamelAnnotator implements Annotator {
 
+    protected final IdeaUtils ideaUtils = IdeaUtils.getService();
+    protected final CamelIdeaUtils camelIdeaUtils = CamelIdeaUtils.getService();
```
essobedo, line 44: Why do we need fields here instead of calling directly the methods?

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/annotator/CamelBeanMethodAnnotator.java
```diff
@@ -112,26 +116,11 @@ public void annotate(@NotNull PsiElement element, @NotNull AnnotationHolder hold
 
     @NotNull
     public List<PsiMethod> getMatchingMethods(PsiClass psiClass, String methodName) {
-        return getJavaMethodUtils().getBeanMethods(getJavaMethodUtils().getMethods(psiClass))
+        final JavaMethodUtils javaMethodUtils = JavaMethodUtils.getService();
+        return javaMethodUtils.getBeanMethods(javaMethodUtils.getMethods(psiClass))
```
essobedo, line 120: Why adding a field instead of calling directly the method?

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/annotator/CamelJSonPathAnnotator.java
```diff
@@ -53,7 +49,7 @@ boolean isEnabled() {
     void validateText(@NotNull PsiElement element, @NotNull AnnotationHolder holder, @NotNull String text) {
 
         // only validate if the element is jsonpath element
-        if (getCamelIdeaUtils().isCamelExpression(element, "jsonpath")) {
+        if (camelIdeaUtils.isCamelExpression(element, "jsonpath")) {
```
essobedo, line 52: I prefer that we call the method instead of having to deal with fields as it also means a state

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/completion/extension/CamelJavaBeanReferenceSmartCompletion.java
```diff
@@ -49,14 +50,15 @@ public class CamelJavaBeanReferenceSmartCompletion extends CompletionProvider<Co
     @Override
     protected void addCompletions(@NotNull CompletionParameters completionParameters, @NotNull ProcessingContext processingContext, @NotNull CompletionResultSet completionResultSet) {
         final PsiElement element = completionParameters.getPosition();
-        final PsiClass psiClass = getCamelIdeaUtils().getBean(element);
+        final PsiClass psiClass = CamelIdeaUtils.getService().getBean(element);
+        final JavaMethodUtils javaMethodUtils = JavaMethodUtils.getService();
```
essobedo, line 54: Only needed in the `if` block

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/YamlCamelIdeaUtils.java
```diff
@@ -75,7 +74,7 @@ public class YamlCamelIdeaUtils extends CamelIdeaUtils implements CamelIdeaUtils
      * All keys whose value can be a URI.
      */
     private static final String[] PLACE_FOR_ENDPOINT_URI = {"uri", "dead-letter-uri", "from", "to"};
-
+
```
essobedo, line 77: Not needed

essobedo (commented):
Great work! it is definitively much better. Only few more comments/questions.

camel-idea-plugin/src/main/java/com/github/cameltooling/idea/annotator/CamelBeanMethodAnnotator.java
ammachado: It's used twice on line 120.

camel-idea-plugin/src/main/resources/META-INF/plugin.xml
ammachado: `CamelPluginStartup` was using a deprecated method (`projectOpened`).

essobedo:
There are some compilation issues

essobedo:
We are good to go, thank you

Issue #820, opened 2023-04-13
Is debugging in Camel Quarkus supported?


javaduke:
I have a Camel Quarkus application (in XML DSL) which I'm trying to debug, but when I run it, all breakpoints that I set in the code are disabled. I use Camel Quarkus Application runner, do I need to specify any additional parameters or profiles?
Forgot to mention, I use the latest version of the plugin, 0.9.4 and IntelliJ 2023.1

essobedo:
IIRC, it should work since https://github.com/camel-tooling/camel-idea-plugin/pull/728. Do you have a reproducer to share?

javaduke:
No, sorry, it's a proprietary code and I cannot share it. But I have several projects based on different versions of Camel (3.18, 3.20, etc) and they all show the same behavior. Again, the code is written using XML DSL, not Java DSL, maybe this is an issue?

javaduke:
OK, I think I know what's happening. Even though I see that the Camel debugger is enabled, the IntelliJ still starts Ava debugger on a different port. In the console log I see: `Listening for transport dt_socket at address: 5005. Attach debugger` but even when I click on Attach debugger and it connects to port 5005, it switches back to the Java debugger.
Not sure how to address this.

javaduke:
Oddly enough it doesn't stop at the Java breakpoints either (although it still shows them as Enabled). Is there anything specific I should have in my pom or application.yml? I use the Camel Quarkus Application run configuration, tried explicitly setting the profile `dev`, but nothing helps. Something is going on here...

javaduke:
And here's what I see in the Breakpoints window:
![Screenshot 2023-04-13 at 1 48 27 PM](https://user-images.githubusercontent.com/561559/231879559-87d9ee74-c440-4d9e-a758-9da4c82df5c0.png)

javaduke:
I ran the code in the debugger and what happens is the result of the xpath evaluate() is null, see here: https://github.com/camel-tooling/camel-idea-plugin/blob/main/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java#L829

The path expression seems to be correct, e.g.:
`//*[@sourceLocation='file:/Users/eberman/Dev/demobank/target/classes/camel/accounts.xml' and @sourceLineNumber='10']` but the result is null.

javaduke:
https://github.com/camel-tooling/camel-idea-plugin/pull/821

[...]

PR #821, opened 2023-04-14
Fix for issue #820 - resolve source location in the classpath

javaduke:


camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java
```diff
@@ -802,7 +802,8 @@ private String getBreakpointId(@NotNull PsiElement breakpointTag) {
             final String url = virtualFile.getPresentableUrl();
             if (virtualFile.isInLocalFileSystem()) { //TODO - we need a better way to match source to target
                 sourceLocations = List.of(
-                    String.format("file:%s", url.replace("src/main/resources", "target/classes")) // file:/absolute/path/to/file.xml
+                    String.format("file:%s", url.replace("src/main/resources", "target/classes")),// file:/absolute/path/to/file.xml
+                    String.format("classpath:%s", url.substring(url.lastIndexOf("src/main/resources/") + 19 ))
```
essobedo, line 806: Could you please clarify a bit more by adding a comment

javaduke: According to my observations, Camel and Camel SpringBoot run configurations pull the code from `target/classes`, but Camel Quarkus builds the JAR first and then runs the JAR. So when the debugger calls Camel and requests XML routes dump, the `sourceLocation` attribute value is in form of `classpath:myroutesfile.xml` as opposed to `file:/com/foo/bar/target/classes/myroutesfile.xml`. Since the plugin is generic, i.e. it should support Camel, SpringBoot and Quarkus run configurations, I simply add `classpath:...` to the list of source locations.

essobedo: Would you mind adding this info as comment into the code?

javaduke: done.

javaduke: I found one more related issue, the step over and step into don't work, the problem is here:
https://github.com/camel-tooling/camel-idea-plugin/blob/main/camel-idea-plugin/src/main/java/com/github/cameltooling/idea/runner/debugger/CamelDebuggerSession.java#L862
But I don't know yet how to address it...

essobedo: I planned to work on the plugin next week, if you can wait a bit, I'll look into that too

javaduke: Actually I fixed it too, see my last commit. It is essentially the same problem, classpath vs absolute path.

I feel like the whole debugger code is a bit clunky and could  use some refactoring and refinement, but unfortunately I don't have any time to work on it at least for the next two-three months...

davsclaus (approved):

PR #840, opened 2023-05-15
Ref #641: Make Camel Debugger support Gradle

### LINKED ISSUE #641: Support debugger for non-Maven based projects
Current implementation of the debugger implies that the projects are Maven-based. We need to add support for other types of projects, Gradle-based, for example.

essobedo:
fixes #641 

## Motivation

The current implementation of the debugger implies that the projects are Maven-based. We need to add support for other types of projects, Gradle-based, for example.

## Modifications:

* To avoid affecting projects for nothing, the Camel debugger is only enabled when at least one Camel breakpoint has been added
* Ensures that a warning is displayed when something when wrong during the configuration of the Camel Debugger
* Auto-generates a grade build file (Groovy or Kotlin) to add the camel debugger artifacts in case of a Grade task
* Auto-configures the launcher in case of a Grade task or a main class
* Moves the type of Camel Debugger Context into a dedicated enumeration
* Forks the code of `ExternalSystemTaskDebugRunner` and `CamelForkedDebuggerThread` to be able to build a `ContextAwareDebugProcess`

## Result

In the case of a generic Gradle task, the plugin is able to add and configure the Camel Debugger by itself as shown below:

46-camel-debugger-gradle-springboot-java-test-launcher

However, in the case of a main class, the plugin can auto-configure it but it cannot add the Camel Debugger automatically, a warning is shown when it happens.


davsclaus (approved):

davsclaus:
Really good work yet again. Can you update the changelog in the plugin.xml file also

essobedo:
@davsclaus Do you mean this https://github.com/camel-tooling/camel-idea-plugin/pull/840/files#diff-98d2d73161e7f3e7be4baa3bd7d326019e986123208137a89ae8dbb2b894dd8aR15?

davsclaus:
Ah thanks I missed that you have done that

PR #845, opened 2023-05-17
Ref #691: Allow to enable manually the code assistance

### LINKED ISSUE #691: camel k / camel-jbang - Add support for code assistance
We should be able to load a file (yaml, java, xml) etc. into IDEA without maven/gradle and then have code assistance for Apache Camel.

This can be done in VSCode today with the Camel plugin there.

Otavio has build a prototype for IDEA that can do some of that.

It would be good to maybe see if we can have this in this IDEA plugin so we have it all together.

essobedo:
fixes #691 

## Motivation

We should be able to load a file (yaml, java, xml) etc. into IDEA without maven/gradle and then have code assistance for Apache Camel.

## Modifications:

* Adds the ability to indicate manually that the project is a Camel project thanks to the project's settings
* Adds the ability to indicate manually the version of the Camel catalog to load thanks to the project's settings
* Ensures that the version of the catalog corresponding to the project settings is loaded at project startup and at the project's settings change

## Result

Thanks to the manual configuration, it is now possible to have access to the code assistance for camel-k / camel-jbang files

 
49-camel-jbang-code-assistance


davsclaus (approved):

davsclaus:
yay this is fantastic, love that we have our first support for code assistance with camel-jbang now

Issue #856, opened 2023-05-31
Compatibility problem with intellij idea version 2023.1


humber42:
There is a problem with the compatibility of the Apache Camel plugin with the version of IntelliJ IDEA v2023.1 which does not allow it to be installed

davsclaus:
Yeah I can install 0.9.9 on 2023.1.2 but then when I open a Camel example the plugin is grey

davsclaus:
You can try with 0.9.8 version which you can find here on github and download and install manually

davsclaus:
```
com.intellij.diagnostic.PluginException: class com.intellij.psi.impl.source.JavaDummyHolder cannot be cast to class com.intellij.psi.PsiJavaFile (com.intellij.psi.impl.source.JavaDummyHolder and com.intellij.psi.PsiJavaFile are in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @5d366db2) [Plugin: org.apache.camel]
	at com.intellij.diagnostic.PluginProblemReporterImpl.createPluginExceptionByClass(PluginProblemReporterImpl.java:23)
	at com.intellij.diagnostic.PluginException.createByClass(PluginException.java:99)
	at com.intellij.psi.impl.source.codeStyle.CoreCodeStyleUtil.processElementOrFail(CoreCodeStyleUtil.java:86)
	at com.intellij.psi.impl.source.codeStyle.CoreCodeStyleUtil.postProcessElement(CoreCodeStyleUtil.java:59)
	at com.intellij.formatting.service.CoreFormattingService.formatElement(CoreFormattingService.java:45)
	at com.intellij.formatting.service.FormattingServiceUtil.formatElement(FormattingServiceUtil.java:67)
	at com.intellij.psi.impl.source.codeStyle.CodeStyleManagerImpl.reformat(CodeStyleManagerImpl.java:81)
	at com.intellij.psi.impl.source.codeStyle.CodeStyleManagerImpl.reformat(CodeStyleManagerImpl.java:64)
	at com.intellij.codeInsight.daemon.impl.quickfix.AddTypeCastFix.createCastExpression(AddTypeCastFix.java:109)
	at com.intellij.codeInsight.daemon.impl.quickfix.CastMethodArgumentFix$MyFixerActionFactory.getModifiedArgument(CastMethodArgumentFix.java:45)
[...]

PR #859, opened 2023-06-03
#856 - Fix NPE

davsclaus:


camel-idea-plugin/src/main/java/com/github/cameltooling/idea/service/extension/camel/JavaCamelIdeaUtils.java
```diff
@@ -116,7 +116,8 @@ public class JavaCamelIdeaUtils extends CamelIdeaUtils implements CamelIdeaUtils
 
     @Override
     public boolean isCamelFile(PsiFile file) {
-        if (file != null && JavaFileType.INSTANCE.equals(file.getFileType())) {
+        if (file != null && JavaFileType.INSTANCE.equals(file.getFileType())
+            && file instanceof PsiJavaFile) {
             PsiJavaFile javaFile = (PsiJavaFile) file;
```
essobedo, line 121:
```suggestion
            && file instanceof PsiJavaFile javaFile) {
```

essobedo:
The failures are due to https://github.com/camel-tooling/camel-idea-plugin/pull/857

davsclaus:
Can we avoid that kubernetes JAR ; what do we really use it for

essobedo:
It is needed for this https://github.com/camel-tooling/camel-idea-plugin/pull/709

essobedo:
@davsclaus I confirm that I get this error very regularly so it is quite annoying. I'm wondering if we could have a new release with this bug fix, WDYT?

davsclaus:
yeah sure

Issue #883, opened 2023-07-05
Upgrade to Camel 3.21.1


davsclaus:

davsclaus:
Okay so the 3.21 upgrade likely needs some adjustments in the camel idea plugin, could be related to the camel-catalog-maven

davsclaus:
This will require
https://issues.apache.org/jira/browse/CAMEL-19580

And therefore 3.21.1

davsclaus:
We will upgrade to Camel 4 instead

PR #886, opened 2023-07-05
#883: Upgrade to Camel 3.21.0

davsclaus:


davsclaus:
We need 3.21.1 or newer

PR #907, opened 2023-09-27
Ref #898: Upgrade to Camel 4.0

### LINKED ISSUE #898: Upgrade to Camel 4.0

essobedo:
fixes #898 

## Motivation

Camel 4.0 has been released so we would like to upgrade Camel to that version.

## Modifications:

* Use a specific version for the Karaf catalog as there is no version 4 available
* Add Groovy/Ivy as runtime dependencies explicitly since they are needed by the plugin since they are no longer transitively inherited by Camel
* Adapt the code to Camel 4
* Adapt some tests to the Camel 4 catalog
* Remove the test ensuring the deprecating warning could be detected by the plugin because there are no more deprecated options in the Camel 4 catalog
* Upgrade all camel dependencies to the latest versions


davsclaus (approved):
Oh that is nice to see that the migration was not very big.
Yes the maven downloaded is another task for another PR ;)

PR #915, opened 2023-10-02
Ref #850: Add a Camel JBang dependency loader

### LINKED ISSUE #850: Camel-JBang - Auto-import dependencies
When using Camel-JBang, if we want to write a route using the Java DSL, for now we need to add manually the dependencies otherwise the class won't compile and the plugin won't allow adding "Camel Breakpoints"

essobedo:
fixes #850 

## Motivation

When using Camel-JBang, if we want to write a route using the Java DSL, for now, we need to manually add the dependencies otherwise the class won't compile and the plugin won't allow adding "Camel Breakpoints".

## Modifications:

* Add a new configuration entry in `Settings.../Languages & Frameworks/Apache Camel/Miscellaneous` to indicate explicitly that a project is a JBang project
* Add a new item menu in the Tools menu that appears only if it detects the folder `.camel-jbang` or the project has explicitly been configured in `Settings.../Languages & Frameworks/Apache Camel/Miscellaneous` as a Camel and JBang project
* Disable the code allowing to set a random debug port until https://github.com/jbangdev/jbang/issues/1689 is fixed
* Improve the way to handle exceptions when the Camel JBang Debugger cannot be launched
* Add the ability to load the camel version from the file `camel-jbang-run.properties`
* Improve the way to define the source locations to better support Java classes when debugging a route written with the Java DSL
* Add a new service `CamelJBangService` that is able to download and add all the Camel JBang dependencies to the project.

## Result

53-camel-jbang-dependency-loader


davsclaus (approved):

davsclaus:
Great work, nice to see the video of this in action

PR #970, opened 2023-12-08
IntelliJ 2023.3 support

ammachado:
IntelliJ 2023.3 support


essobedo:
Thx for the PR, but it is a little bit more complex than that since we support only 3 minor versions of IDEA so it is a sliding window. I'm closing it for https://github.com/camel-tooling/camel-idea-plugin/pull/971

PR #971, opened 2023-12-09
Upgrade to Idea 2023.3.2

essobedo:
Fixes https://github.com/camel-tooling/camel-idea-plugin/issues/972

## Motivation

A new maintenance version of Idea 2023.3 has been released, so the version of Idea used should be upgraded consequently.

 ## Modifications

* Update the default version of Idea used
* Update the list of versions of Idea to test
* Update the readme accordingly


davsclaus (approved):

davsclaus:
> Cannot find builtin plugin 'xpath' for IDE: /home/runner/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/2023.3/6105b81c6142f62379ad6c5afb542c77350a71eb/ideaIC-2023.3

I wonder if they removed some plugins in that never version

javaduke:
I could be wrong but I think the "xpath" module can be safely removed.

essobedo:
It should be released in 2 weeks

Issue #985, opened 2024-02-12
Upgrade to Camel 4.4


davsclaus:
Will be released next week

PR #987, opened 2024-02-19
#985: Upgrade to Camel 4.4.0

davsclaus:


essobedo (approved):
LGTM

gradle.properties
```diff
@@ -1,6 +1,6 @@
-camelVersion = 4.3.0
-camelQuarkusVersion = 3.6.0
+camelVersion = 4.4.0
+camelQuarkusVersion = 3.7.3
```
essobedo, line 2: It looks like the latest version of the quarkus catalog is still 3.7.0

davsclaus: Thanks I just spotted that too.

PR #1012, opened 2024-04-10
Upgrade to IntelliJ 2024.1

ammachado:
And remove the upper bound version limit for IntelliJ


davsclaus:
Yeah its a bit weird but jetbrains seems to be using `2024.1` as version instead of `2024.1.0`

davsclaus:
I was able to update gradle in another isolated PR
https://github.com/camel-tooling/camel-idea-plugin/pull/1014

And there is another PR for the IDEA upgrade but it fails also
https://github.com/camel-tooling/camel-idea-plugin/pull/1013

davsclaus:
Thanks @ammachado for the help - we have upgraded with the help from your PR

Issue #1011, opened 2024-04-04
Upgrade to IDEA 2024.1


davsclaus:
So we can support this new version also

davsclaus:
Its all people doing work in their sparetime on weekends and whatnot - you cannot ever expect quick and speedy support. Instead you can help instead by working on the plugin and send PRs.

And Jetbrains tend to change their IDE between new big releases that causes upgrade pain - and this new 2024 version seems to be yet another one.

davsclaus:
okay we found and have fixed this upgrade

PR #1013, opened 2024-04-14
#1011 - Upgrade to IDEA 2024.1

davsclaus:


davsclaus:
@ammachado @essobedo okay so this can work locally for me on mac osx. But the CI job seems to fail on ubuntu and I can't see why.

davsclaus:
If I install the plugin from disk then it works on my osx. I just think its some weird with that CI build and testing. I can't see the test logs/error

davsclaus:
okay I got it fixed now, was some tests that needed to change to another base class and a few that had some mock jdk set for 11 instead of 17

PR #1041, opened 2024-08-12
20242

davsclaus:


davsclaus:
Okay @essobedo so the 2024.2 IDEA is Java 21 only, and its likely to hard to maintain older versions.
So we can bump the version from 1.2 to 1.3, and then do 2024.2 onwards. But on top of that then IDEA has other changes in the plugin dev kit so something more have to change in gradle build files and whatnot --- argh its not fun

essobedo:
Closing it as it has been replaced by https://github.com/camel-tooling/camel-idea-plugin/pull/1048

PR #1073, opened 2024-11-15
"Open-End Compatibility", take 2

ammachado:


build.gradle
```diff
@@ -142,11 +117,26 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
 }
 
-description = 'Apache Camel IDE :: IDEA Plugin'
+intellijPlatform {
+    pluginVerification {
+        ides {
+            recommended()
```
essobedo, line 123: IMO, I'm not really a fan of losing the list of versions officially supported (reason why I didn't do it during the migration), it is already complex to support the 4 latest versions if we no longer keep control of the list, it will be even worse. 

@davsclaus WDYT?

davsclaus:
Hmm there is a new 2024.3 release now and this plugin would then need to be updated and release to support this. Until then the only version users see is a very old 0.5.x release.

The open end allows users to upgrade and still use the plugin (it may not work) but then they can report it here, and we can still do the tests on X idea versions as we do now (I like that). But IMHO its annoying to have to keep releaseing this plugin whenvever Jetbrains do a new release.

I wonder if we can have the existing testing of those versions as before, and make the "until" more open so the plugin can be installed on newer versions

PR #1126, opened 2025-07-19
support for intellij 2025 - fix tests

rastislavpapp:
I see that you've already implemented the support for 2025.1, but when I checked it out the tests did not pass, I've tried to fix them.

Also, on Windows they do not pass, because of a problem with classgraph vs intellij's MultiRoutingFileSystem - it's somehow set as a default filesystem, and when classgraph tries to convert Path to File, at `nonapi.io.github.classgraph.classpath.ClasspathOrder#addClasspathEntry`, it fails because `getFileSystem() == FileSystems.getDefault()` returns false. I can take a look at it as well, this might maybe be an issue at runtime (on windows)?


davsclaus:
Thanks for all the help - you are very much welcome to continue so we can get this working on latest IDEA. And yeah if you can find and fix the windows problem then that is awesome - there is a lot of windows users.

PR #1128, opened 2025-07-19
Update to IDEA 2025

davsclaus:
This would drop support for older releases as its too painful to have backwards support. But the older releases works on IDEA 2024 and 2023.


davsclaus:
Yay I can build a .zip locally and install this to IDEA 2025 from the file.

rastislavpapp:
the tests pass! hurray!

PR #1129, opened 2025-07-19
fix CI actions and more windows tests

rastislavpapp:
i've updated the github actions to work correctly, and this revealed one more problem with tests on windows:

* the github test environment for a windows container puts the cloned repository into D:/a directory, which is not under user $HOME. This causes a problem with failing `VfsRootAccess#assertAccessInTests`, the tests are touching files they should not be touching. By default, user home dir is included in the allowed directories, and this is why it works on linux and macos (and my windows notebook :) ), but when the project repository is cloned outside $HOME, and tests start to access src/test/resources/testData, the `VfsRootAccess#assertAccessInTests` does not like it.
* fixed by including the root of the testData dir in allowed dirs. I wonder why this is necessary, as being accessed is the main purpose of test files; we are probably doing something wrong in our test setup, but I don't have the necessary plugin development knowledge to find out what it is. The documentation seems contradictory, for example, `LightJavaCodeInsightFixtureTestCase#getTestDataPath`'s javadoc says the method should not be overriden (which we do), but all the examples I found override it :) 
* so I just included the test dir manually.


davsclaus (approved):

davsclaus:
Thanks a lot for all this awesome help. I wonder if you would like to be a member of the team. I have sent you an invite via github

rastislavpapp:
Sure, thank you :-)

PR #1137, opened 2025-08-02
many improvements / fixes regarding property placeholders, endpoint references and line markers

rastislavpapp:


davsclaus:
No problem I know the feeling and do that myself also. Really appreciate that you step up and help with this tool. Thanks for adding new tests also for the YAML DSL and so on. 

We also have some warnings from IDEA about using deprecated APIs and functionality that can change or be removed in the future. So there is also some cleanup work there too. Not sure if the gradle plugin can do this report, but when I upload a new version to IDEA they generate a report afterwards.

rastislavpapp:
sorry about that, I've commited the license headers :)

davsclaus:
LGTM you are welcome to merge.

Also the work you do you are welcome to add a note in plugin.xml so its shown when users update

rastislavpapp:
I've added change notes, merging now.

PR #1142, opened 2025-08-10
support for property references in the value of blueprint's ConfigProperty annotation

rastislavpapp:


davsclaus (approved):

rastislavpapp:
argh, I'll get used to it :D eventually

PR #1154, opened 2025-08-18
support blueprint's factory-method and factory-ref

rastislavpapp:
References / code completion for blueprint bean's factory-ref and factory-method. Expanding on current functionality for bean references. 
Also helps with `@BeanInject` type validation, until now, beans created via factories had an unknown type. With this enhancement, the types of beans created via factories can be determined, so there's less warnings, or possibly errors for incompatible bean types. (As I'm writing this, I'm realizing that I forgot to add tests for this, I'll add some)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
-camelVersion = 4.13.0
-camelQuarkusVersion = 3.24.0
+camelVersion = 4.14.0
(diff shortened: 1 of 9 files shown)


davsclaus:
LGTM

davsclaus:
is there more work to this

As we got 4.14 released then I think its good to get a new release out

PR #1164, opened 2025-08-20
recognize endpoint references for @Consume, fixes #921

### LINKED ISSUE #921: Plugin does not recognize @Consume endpoints
IntelliJ keeps telling me that my direct endpoint does not exist when using the .to() component.

This will show an error marker "Cannot find endpoint declaration":
```
@Consume("direct:target")
public void doSomething() {}

public void test() {
  from("direct:source").to("direct:target")
}
```

Do we have a workaround for this or is it the expected behaviour?

rastislavpapp:


davsclaus:
can you update the changelog in the plugin.xml

PR #1166, opened 2025-08-21
support blueprint property placeholders

rastislavpapp:
property placeholders instyle of `${...}`, in blueprint xml files

### DIFF
--- a/src/main/java/com/github/cameltooling/idea/Constants.java
+++ b/src/main/java/com/github/cameltooling/idea/Constants.java
@@ -20,9 +20,10 @@ public final class Constants {
 
     public static final String OSGI_BLUEPRINT_NAMESPACE = "http://www.osgi.org/xmlns/blueprint/v1.0.0";
     public static final String CAMEL_NAMESPACE = "http://camel.apache.org/schema/spring";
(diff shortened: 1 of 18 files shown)


davsclaus (approved):

rastislavpapp:
> And also remember to update the changelog in plugin.xml

I've managed not to forget this time :-)

> This style with ${ } is also the same for classic Spring <beans> XML files. Would you be able to add that also in a new PR?

This should already be handled by spring plugin itself. Which is part of Ultimate, and I don't know how would Jetbrains react to us replicating ultimate features in a CE plugin :-)
I've actually thought about whether the blueprint support should be a part of this camel plugin or should be split into a separate blueprint plugin, because it does not really have anything to do with camel :-) but there's _some_ overlap (e.g. knowing about blueprint beans we can interact with them in `@BeanInject` etc. But it could be possible to have some kind of inter-plugin communication, or awareness.

davsclaus:
osgi blueprint is dead - its only camel karaf users that use it - so it doe not make sense to build new plugins

you can add a [x] in settings to turn this on | off then users can enable this that do not use ultimate version (which is also a lot of users)

PR #1177, opened 2025-08-31
add references from endpoint uri parameters to their corresponding setters in camel classes (*Endpoint, *Configuration)

rastislavpapp:
like so:

image

pointing to `org.apache.camel.component.file.GenericFileEndpoint#setAppendChars`

I'm not quite sure how to determine which class the query parameter is actually located in, and whether I can determine this programmatically, so I went with the most primitive approach - take the component class name (e.g. DirectComponent), replace Component with Endpoint, or Configuration, and try to find appropriate setter there. If there's a better way, or if there's more possible places to look in, I'll gladly change the logic.

### DIFF
--- a/src/main/java/com/github/cameltooling/idea/reference/endpoint/parameter/CamelEndpointParameterReferenceContributor.java
+++ b/src/main/java/com/github/cameltooling/idea/reference/endpoint/parameter/CamelEndpointParameterReferenceContributor.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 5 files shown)


davsclaus:
what is the purpose of this ? The tooling should not directly use any configurer classes. The information about which options a camel component / endpoint supports is already defined in json files from camel-catalog, or from 3rd party components inside their JARs as json files.

rastislavpapp:
It's so that you can click on a parameter name in an endpoin uri and it'll take you to the actual setter in the appropriate class, like in the example above.
As a user I find myself looking for this kind of information all the time, to look in the javadoc, or actually see where the parameter is used and for what.

davsclaus:
Ah okay, yeah that makes it quicker to dive into the camel source code

PR #1236, opened 2026-04-08
Upgrade to Camel 4.18.1

davsclaus:


rastislavpapp:
Ah, the good old `VfsRootAccess$VfsRootAccessNotAllowedError: ` :-) I've encountered this many times. Tests on windows are actually a bit more strict. Or, maybe I should say - tests on unix are a little bit broken.

IntelliJ test framework has some expectations about which files/directories should be accessible by the tests, and throws this exception when accessing the files outside those. One of the allowed dirs is `user.home`, and on unix agents the whole intellij test instance is placed here. So it circumvents the checks completely.

On windows, the test runs in a root directory under C:\ (I think C:\work or something similar), which is not included, so the verification actually does something. 

I think the problem is that this check is not very deterministic; often there's some completely unrelated test failing after unrelated changes. In this case, the plugin is locating the camel-kamelets*.jar in the plugin install directory, which is not allowed, so it should have been failing from the start. I think.

rastislavpapp:
I've fixed this in #1238

davsclaus:
thanks closing this

